PR #24, opened 2018-03-21
Fix for OSX compile errors

dfrencham:
Fix for g++ compile errors under OSX.

`/Library/Ruby/Gems/2.3.0/gems/arduino_ci-0.1.7/cpp/arduino/Print.h:57:86: error: unknown type name 'PGM_P'`
`/Library/Ruby/Gems/2.3.0/gems/arduino_ci-0.1.7/cpp/arduino/WString.h:39:16: error: no member named 'isnan' in namespace 'std'`

### DIFF
--- a/cpp/arduino/Arduino.h
+++ b/cpp/arduino/Arduino.h
@@ -73,6 +73,7 @@ inline unsigned int makeWord(unsigned char h, unsigned char l) { return (h << 8)
 
 #define word(...) makeWord(__VA_ARGS__)
 
-
-
-
+// audio 
+inline void tone(uint8_t _pin, unsigned int frequency, unsigned long duration) { }
+inline void tone(uint8_t _pin, unsigned int frequency) { }
+inline void noTone(uint8_t _pin) { }
--- a/cpp/arduino/Arduino.h
+++ b/cpp/arduino/Arduino.h
@@ -74,6 +74,7 @@ inline unsigned int makeWord(unsigned char h, unsigned char l) { return (h << 8)
 #define word(...) makeWord(__VA_ARGS__)
 
 // audio 
+// TODO Add audio to GODMODE mock
 inline void tone(uint8_t _pin, unsigned int frequency, unsigned long duration) { }
 inline void tone(uint8_t _pin, unsigned int frequency) { }
 inline void noTone(uint8_t _pin) { }
--- a/cpp/arduino/ArduinoDefines.h
+++ b/cpp/arduino/ArduinoDefines.h
@@ -89,3 +89,6 @@
 #define TIMER5B 17
 #define TIMER5C 18
 
+#if defined(__AVR_ATmega328P__) || defined(__AVR_ATmega32U4__) || defined(__AVR_ATmega328__) || defined(__AVR_ATmega168__) || defined(__AVR_ATmega1280__) || defined(__AVR_ATmega2560__)
+#define LED_BUILTIN 13
+#endif
--- a/cpp/arduino/HardwareSerial.h
+++ b/cpp/arduino/HardwareSerial.h
@@ -61,7 +61,7 @@ class HardwareSerial : public Stream, public ObservableDataStream
     inline size_t write(long n) { return write((uint8_t)n); }
     inline size_t write(unsigned int n) { return write((uint8_t)n); }
     inline size_t write(int n) { return write((uint8_t)n); }
-    // using Print::write; // pull in write(str) and write(buf, size) from Print
+    using Print::write; // pull in write(str) and write(buf, size) from Print
     operator bool() { return true; }
 
 };
--- a/cpp/arduino/Print.h
+++ b/cpp/arduino/Print.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <stdio.h>
+#include <avr/pgmspace.h>
 #include "WString.h"
 
 #define DEC 10
--- a/cpp/arduino/WString.h
+++ b/cpp/arduino/WString.h
@@ -4,6 +4,7 @@
 #include <string.h>
 #include <algorithm>
 #include <iostream>
+#include <cmath>
 #include "AvrMath.h"
 #include "WCharacter.h"
 


ifreecarve:
Can you post your g++ version?  Also, what were you compiling that caused this error?  I"m having some issues now between g++-4 and g++-7, and not sure how best to resolve them.

dfrencham:
I ran _arduino_ci_remote.rb_ to run my tests locally.

g++ on OSX appears to be a link to clang, and isn't g++ at all.

```
danny@stevejerbs:~/Documents/Arduino/rad-gate$ arduino_ci_remote.rb
Installing library under test...                                               ✓
Library installed at /Users/danny/Documents/Arduino/libraries/rad-gate...      ✓
Checking GCC version... 
    Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
    Apple LLVM version 9.0.0 (clang-900.0.39.2)
    Target: x86_64-apple-darwin17.3.0
    Thread model: posix
    InstalledDir: /Library/Developer/CommandLineTools/usr/bin
...Checking GCC version                                                        ✓
Switching to board for uno (arduino:avr:uno)...                                ✓
Unit testing test1.cpp... 
```

dfrencham:
Here is the full compile command being called:

`Last command:  $ g++ -std=c++0x -o /Users/danny/Documents/Arduino/rad-gate/unittest_test1.cpp.bin -DARDUINO=100 -D__AVR_ATmega328P__ -I/Library/Ruby/Gems/2.3.0/gems/arduino_ci-0.1.7/cpp/arduino -I/Library/Ruby/Gems/2.3.0/gems/arduino_ci-0.1.7/cpp/unittest -I/Users/danny/Documents/Arduino/rad-gate -I/Users/danny/Documents/Arduino/rad-gate/libs/Adafruit_NeoPixel -I/Users/danny/Documents/Arduino/rad-gate/libs/Adafruit_Soundboard -I/Users/danny/Documents/Arduino/rad-gate/libs/JQ6500_Serial /Library/Ruby/Gems/2.3.0/gems/arduino_ci-0.1.7/cpp/arduino/Arduino.cpp /Library/Ruby/Gems/2.3.0/gems/arduino_ci-0.1.7/cpp/arduino/Godmode.cpp /Library/Ruby/Gems/2.3.0/gems/arduino_ci-0.1.7/cpp/unittest/ArduinoUnitTests.cpp /Users/danny/Documents/Arduino/rad-gate/AudioFX.cpp /Users/danny/Documents/Arduino/rad-gate/Gate.cpp /Users/danny/Documents/Arduino/rad-gate/LightTree.cpp /Users/danny/Documents/Arduino/rad-gate/Sequence.cpp /Users/danny/Documents/Arduino/rad-gate/constants.cpp /Users/danny/Documents/Arduino/rad-gate/libs/Adafruit_NeoPixel/Adafruit_NeoPixel.cpp /Users/danny/Documents/Arduino/rad-gate/libs/Adafruit_NeoPixel/esp8266.c /Users/danny/Documents/Arduino/rad-gate/libs/Adafruit_Soundboard/Adafruit_Soundboard.cpp /Users/danny/Documents/Arduino/rad-gate/libs/JQ6500_Serial/JQ6500_Serial.cpp /Users/danny/Documents/Arduino/rad-gate/utility.cpp /Users/danny/Documents/Arduino/rad-gate/test/test1.cpp`

cpp/arduino/Arduino.h
```diff
@@ -73,6 +73,7 @@ inline unsigned int makeWord(unsigned char h, unsigned char l) { return (h << 8)
 
 #define word(...) makeWord(__VA_ARGS__)
 
-
-
-
+// audio 
+inline void tone(uint8_t _pin, unsigned int frequency, unsigned long duration) { }
```
ifreecarve: please add a `TODO` comment here, so that I remember to add these to the `GODMODE` mock

ifreecarve:
Thanks. Sorry for the delay on this, I've been on parental leave.

ifreecarve:
I want to fix this stuff more, but I don't want to delay getting these fixes out either.  As a compromise I put together #28 (branch is `2018-03-27_compilation`) -- it incorporates your changes but puts the edits in the files where the proper solutions will eventually live.

Would you mind giving that branch a try and seeing if it fixes the errors you saw?  I tested locally with `clang`, `g++-4.7`, and `g++-7` and all seem to work.

dfrencham:
I tried `2018-03-27_compilation` but discovered I hadn't committed a change to HardwareSerial. Once I applied that, it worked fine. I noticed you had that line commented out - any idea why?

ifreecarve:
Just to be clear, I'm going to merge #28 and close this request -- but only after I make sure that I have tested and fixed the problems you're seeing here.

I commented out `Print::write` from `HardwareSerial` because it should be pulling in the `Stream::write` instead.  Basically I need it to fall back on my GODMODE mock instead of what would ordinarily be provided by the low-level library.

What did you compile that produced the error with `Print::write`?  I couldn't get it to happen with any of the `SampleProjects/TestSomething/test/*.cpp` files, and I tried `clang`, `g++-4.9`, and `g++-7`.

dfrencham:
I've been able to remove the function that was causing the failure. Regardless, here it is:

```
int aprintf(char *str, ...) {
	int i, j, count = 0;

	va_list argv;
	va_start(argv, str);
	for(i = 0, j = 0; str[i] != '\0'; i++) {
		if (str[i] == '%') {
			count++;
                         
                         // fails here
			Serial.write(reinterpret_cast<const uint8_t*>(str+j), i-j);

			switch (str[++i]) {
				case 'd': Serial.print(va_arg(argv, int));
					break;
				case 'l': Serial.print(va_arg(argv, long));
					break;
				case 'f': Serial.print(va_arg(argv, double));
					break;
				case 'c': Serial.print((char) va_arg(argv, int));
					break;
				case 's': Serial.print(va_arg(argv, char *));
					break;
				case '%': Serial.print("%");
					break;
				default:;
			};

			j = i+1;
		}
	};
	va_end(argv);

	if(i > j) {
                // also fails here
		Serial.write(reinterpret_cast<const uint8_t*>(str+j), i-j);
	}

	return count;
}
```

ifreecarve:
Excellent -- if this code compiles in the Arduino IDE and fails in arduino_ci, then it's "must fix" for me.  I'll check this out.

ifreecarve:
Update: you're right that uncommenting that line was the right thing to do.  On closer inspection, `Serial.print((int) 2)` wasn't working in my implementation either.  I think I've fixed both of those now.

ifreecarve:
Tomorrow I'm going to merge #28 and close this one, unless you discover that the code in #28 doesn't work.  I had more problems than the ones you had found...

But the good news is that you inspired me to add some code where I can specify multiple compilers in local testing, to make sure they all work :)

dfrencham:
Excellent cheers mate. Another feature that would be great would be a configuration parameter to exclude folders from compilation. Up until now I've kept my libraries in ./lib so they can all be in a single git hub repo (using git submodules). With arduino_ci I can't do this, as I noticed it walks the tree to find all the header files - so even if I have compiler directives to use mock headers/classes instead, the libraries get pulled in too (which in my case broke compilation).

If this sounds worth while I can create a bug tagged as "feature request". When I get a chance I can have a crack at it myself, but my knowledge of Ruby wouldn't fill a thimble...

ifreecarve:
I've opened #29 to track that issue, and that sounds like another good priority.  As you noticed, the way I include files at the moment is currently a very blunt instrument.  If you can update that issue with a repo of yours that demonstrates the problem, I can start looking into possible fixes.

PR #34, opened 2018-04-19
Added colors to unit test output

dfrencham:
* console color codes have been added
* changes have been confirmed to work in bash (OSX) terminal, and TravisCI web interface. They will need to be tested in a Windows CMD window

### DIFF
--- a/cpp/unittest/ArduinoUnitTests.h
+++ b/cpp/unittest/ArduinoUnitTests.h
@@ -65,7 +65,7 @@ class Test
               const char* rhsLabel,
               const B &rhs
           ) {
-            cerr << "    " << (pass ? COLOR_GREEN : COLOR_RED) << (pass ? "" : "not ") << "ok " << ++mAssertCounter << " - ";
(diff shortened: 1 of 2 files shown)


ifreecarve:
Whoops, the windows CI shouldn't be running yet -- I haven't implemented it!  I just toggled the setting to ignore branches without an `appveyor.yml` file, so any further pushes you do here should be passed as normal.

ifreecarve:
I like the idea of this, but for cross-platform support I think it would be better to pull in the gem that does this, check out [this answer about the colorize gem](https://stackoverflow.com/a/1489233/2063546).

ianfixes:
I'm going to close this PR for now, but I think the fact that you contributed it speaks to a shortcoming in the current console output that I'd like to address to your satisfaction.

Can you open an issue that describes what aspect of workflow this PR impacts?  Developer ergonomics are the biggest driver of the design of this project, so I want to make sure I'm tackling those to the best of my ability.

PR #52, opened 2018-05-11
Misc improvements

hlovdal:
Two smaller fixes.


ianfixes:
Thanks for finding these!

Issue #13, opened 2018-01-25
libraries aren't included properly during unit test compilation


ianfixes:
Currently `CppLibrary` is using the "base name" of the directory, not the full path.  Need to figure out a way to get the full library installation dir from `ArduinoCmd`.

Further, we are not yet following `#include` statements for any files that might come in from subdirectories.  

There are at least 6 cases that need to be tested, each as its own subdir of `SampleProjects`.

* Arduino 1.0.x compatible where all necessary files are in the root directory
  * https://github.com/kroimon/Arduino-SerialCommand
* Arduino 1.0.x compatible where some necessary files are included from a subdirectory
  * https://github.com/adafruit/Adafruit_FONA
* Arduino 1.0.x compatible where some subdirectories contain files that should _not_ be included
* Arduino 1.5 compatible where all necessary files are in `src/`
* Arduino 1.5 compatible where some necessary files are included from other directories
* Arduino 1.5 compatible where some subdirectories contain files that should _not_ be included

PR #53, opened 2018-05-11
Add limited support for libraries during unit test compilation

hlovdal:
## Issues Fixed

* Partially resolves [issue 13](https://github.com/ianfixes/arduino_ci/issues/13).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+* Libraries installed in `$HOME/Arduino/libraries` and which has its content in a `src` directory are now properly
+included in the compiler command. This partially solves [issue 13](https://github.com/ianfixes/arduino_ci/issues/13).
+
 ### Security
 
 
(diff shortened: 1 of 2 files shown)


ianfixes:
This is a feature I can't accept changes on without unit tests, which in this case would involve new SampleProjects that pull in external libs.  There are at least 6 use cases that need to be supported, so demonstrating any one of these would be an improvement over what I have now:

* Arduino 1.0.x compatible where all necessary files are in the root directory
* Arduino 1.0.x compatible where some necessary files are included from a subdirectory
* Arduino 1.0.x compatible where some subdirectories contain files that should _not_ be included
* Arduino 1.5 compatible where all necessary files are in `src/`
* Arduino 1.5 compatible where some necessary files are included from other directories
* Arduino 1.5 compatible where some subdirectories contain files that should _not_ be included

I'm updating #13 with these cases.

ianfixes:
I'm working on some changes in this area, which might supersede the changes you're making here.  Do you have any example projects that are built using these changes?  I'd like to add tests against those.

ianfixes:
I think #61 covers all of this and more.  If not, please re-open!

PR #71, opened 2018-12-30
Provide missing itoa function

hlovdal:
Provide an `itoa` function. It is present in Arduino's runtime environment but not on most (all?) host systems because itoa is not a portable standard function.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,23 @@ Pull requests will trigger a Travis CI job.  The following two commands will be 
  * `rubocop -D` - code style tests
  * `rspec` - functional tests
 
+ If you do not already have a working ruby development environment set up, run the following commands:
+
+```shell
+apt-get install ruby ruby-dev    # For Debian/Ubuntu
+dnf install ruby ruby-devel      # For Fedora
+yum install ruby ruby-devel      # For Centos/RHEL
+gem install bundler              # See note below about version
+gem install rubocop
+gem install rspec
+```
+
+As of writing this you want install a version 1 of bundler, `gem install bundler -v '1.17.3'`,
+since there is some incompability with regards to the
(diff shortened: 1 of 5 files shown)


ianfixes:
Please add one or 2 unit test assertions for this in `SampleProjects/TestSomething/test/stdlib.cpp`, which will verify that this `itoa` implementation compiles and produces expected results.

ianfixes:
Thanks for this contribution!

CONTRIBUTING.md
```diff
@@ -12,7 +12,23 @@ Pull requests will trigger a Travis CI job.  The following two commands will be
...
+gem install rubocop
+gem install rspec
+```
+
+As of writing this you want install a version 1 of bundler, `gem install bundler -v '1.17.3'`,
+since there is some incompability with regards to the
+[latest version 2 release of bundler](https://bundler.io/blog/2019/01/04/an-update-on-the-bundler-2-release.html)
+
 Be prepared to write tests to accompany any code you would like to see merged.
+See `SampleProjects/TestSomething/test/*.cpp` for the existing tests (run by rspec).
```
ianfixes: Thank you for adding this.  I wouldn't even have thought to ask about what difficulties you encountered from my documentation, so if there's anything else that can be improved I would be glad to write it up.

SampleProjects/TestSomething/test/stdlib.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+unittest(library_tests_itoa)
+{
+  char buf[32];
+  const char *result;
+  struct {
+    int value;
+    const char *expected;
+    int base;
+  } table[] = {
```
ianfixes: TIL this trick.  Nice.

ianfixes:
All of this is excellent.

PR #73, opened 2019-01-02
Bugfixes

ianfixes:

* Check that example builds are actually desired before complaining that none were found
* Fix behavior of config file, where falsey values (empty arrays specifically) did not properly override existing values
* Project directory names can now contain spaces instead of failing cryptically 
* See CHANGELOG.md for more

### DIFF
--- a/exe/arduino_ci_remote.rb
+++ b/exe/arduino_ci_remote.rb
@@ -263,8 +263,11 @@ else
   end
 end
 
-if library_examples.empty?
-  inform_multiline("Skipping libraries; no examples found in #{installed_library_path}") do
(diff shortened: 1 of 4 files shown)


exe/arduino_ci_remote.rb
```diff
@@ -263,8 +263,11 @@ def display_files(pathname)
   end
 end
 
-if library_examples.empty?
-  inform_multiline("Skipping libraries; no examples found in #{installed_library_path}") do
+puts "config.platforms_to_build : #{config.platforms_to_build}"
```
ianfixes: Whoops

PR #95, opened 2019-01-26
Expand pgm_read_... macros

### LINKED ISSUE #74: Fix the signatures of `pgm_read_*` macros
Related to errors like this one

```
error: invalid operands to binary expression ('const uint8_t *' (aka 'const unsigned char *') and 'const uint8_t *')
                crc = pgm_read_byte(dscrc2x16_table + (crc & 0x0f)) ^
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
```

The code in `pgmspace.h` that needs to change is this:
```c++
// this is the one I updated
#define pgm_read_byte(addr) (*(const uint8_t *)(addr))

// these ones need to be updated in similar fashion
#define pgm_read_word(x) (x)
#define pgm_read_dword(x) (x)
#define pgm_read_float(x) (x)
#define pgm_read_ptr(x) (x)
#define pgm_get_far_address(x) (x)
```

The problem is that at the moment I haven't dug into what these signatures should be.

hlovdal:

* The `pgm_read_...` preprocessor macros in `cpp/arduino/avr/pgmspace.h` now expands to an expression with applicable type.

## Issues Fixed

* Fixes #74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Unit tests and examples are now executed alphabetically by filename
+- The `pgm_read_...` preprocessor macros in cpp/arduino/avr/pgmspace.h now expands to an expression with applicable type.
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


hlovdal:
> I assume this fixed a problem for you locally.

Actually no. I was just looking trough the existing issues, and though "This does not look too difficult to fix, let me have a try".

ianfixes:
Oh whoops, I was the one with this problem 🤦‍♂️.  And I even wrote it down... 

https://github.com/ianfixes/arduino_ci/issues/74

Issue #102, opened 2019-01-28
assertEqual does not support nullptr


hlovdal:
Trying to test that a function returns a null pointer, e.g.

```C++
	assertEqual(nullptr, result);
```

does not work

```shell
.../ArduinoUnitTests.h:69:56: error: ambiguous overload for ‘operator<<’ (operand types are ‘std::basic_ostream<char>’ and ‘std::nullptr_t’)
               cerr << "      " << lhsRelevance << ": " << lhs << endl;
```

I tried to fix it by extracting that line into a template function and then make a template specialisation for nullptr but did not get the compiler to select it, and have put that on ice for now since there is a simple workaround, just adding
```C++
#define nullptr (void *)NULL
```
at the top of the test file. But it would be nice to eventually have this fixed.

ianfixes:
Yup, this seems like part of C++11 that needs to be added to the mocks.  What's the data type of `result` in the code you posted?  I'll add that to a unit test.

ianfixes:
I will add `nullptr` and `assertNull(result)`

hlovdal:
Those updates still do not fix the issue with `assertEqual`. The problem is that onAssert tries to print the argument of type `nullptr` which it is not able to. There needs to be an additional unit test:

```C++
unittest(nullpointer_equal)
{
  int* myPointer = NULL;
  int **notNullPointer = &myPointer;
  assertEqual(nullptr, myPointer);
  assertNotEqual(nullptr, notNullPointer);

#if 0
  // The following should not compile, which they currently do not but in a
  // cryptic way. If possible and with low effort an explicit denial which
  // produces an explanatory error would be nice.
  assertLessOrEqual(nullptr, myPointer);
  assertMoreOrEqual(nullptr, notNullPointer);
  assertLess(nullptr, myPointer);
  assertMore(nullptr, notNullPointer);
  assertFalse(nullptr);
  assertTrue(nullptr);
  assertAnyLhsRhsFunction(..., nullptr);
#endif
}
```

ianfixes:
Thanks, those unit tests are exactly what was needed.

ianfixes:
@hlovdal You gave me an excellent idea here.

I've set up a new branch called `tdd` (https://github.com/ianfixes/arduino_ci/tree/tdd).  If you can write unit tests that show a problem, make a PR using the `tdd` branch as a base.  That will give me something to work from when I go about fixing the issue.

ianfixes:
[...]

PR #110, opened 2019-02-03
Correction for issue #102

hlovdal:
Including `Nullptr.h` in `Arduino.h` breaks code that uses `nullptr`. The nullptr define in `Nullptr.h` is only applicable for use for the compare functions and should only be included when compiling the unit test files.

### DIFF
--- a/SampleProjects/DoSomething/do-something.cpp
+++ b/SampleProjects/DoSomething/do-something.cpp
@@ -4,3 +4,19 @@ int doSomething(void) {
   millis();  // this line is only here to test that we're able to refer to the builtins
   return 4;
 };
+
+static const struct something table[] = {
+	{ 1, "abc" },
+	{ 2, "xyz" },
+	{ 4, "123" },
+};
+
+const struct something *findSomething(int id) {
+	for (unsigned int i = 0; i < 3; i++) {
+		if (table[i].id == id) {
+			return &table[i];
+		}
+	}
+	return nullptr;
+}
+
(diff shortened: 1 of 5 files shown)


ianfixes:
I see what I was doing wrong.  After some early experimentation, I had mis-read the error message and assumed that `std::nullptr` couldn't be exposed properly.  I see now that it was a template issue combined with lack of an ostream operator.  

I have different fix for this in https://github.com/ianfixes/arduino_ci/tree/2019-02-04_nullptr that I'm looking at now

hlovdal:
And one more fix:

```patch
    Add missing #include <ostream>

diff --git a/cpp/arduino/Godmode.cpp b/cpp/arduino/Godmode.cpp
index 424dad2..56692fd 100644
--- a/cpp/arduino/Godmode.cpp
+++ b/cpp/arduino/Godmode.cpp
@@ -1,3 +1,5 @@
+#include <ostream>
+
 #include "Godmode.h"
 #include "HardwareSerial.h"
 #include "SPI.h"
diff --git a/cpp/unittest/OstreamHelpers.h b/cpp/unittest/OstreamHelpers.h
index 4bf16c3..a2cd8f4 100644
--- a/cpp/unittest/OstreamHelpers.h
+++ b/cpp/unittest/OstreamHelpers.h
@@ -1,3 +1,5 @@
 #pragma once
 
+#include <ostream>
+
 inline std::ostream& operator << (std::ostream& out, const std::nullptr_t &np) { return out << "nullptr"; }

```

ianfixes:
Can you rebase this off latest master?  I suspect the underlying issue has now been fixed.  If (after merge) the tests I added in #112 still don't seem sufficient, I will be glad to merge this PR to add yours.

ianfixes:
Awesome, I will incorporate that README.

Issue #89, opened 2019-01-14
Add or update board definitions -- arduino mega at least


ianfixes:
Add new board definitions based on [this gist](https://gist.github.com/ah01/762576)

```c++
  #if defined(__AVR_ATmega168__)
    // Arduino Diecimila and older
    LINE("__AVR_ATmega168__", __AVR_ATmega168__);
  #endif
  
  #if defined(__AVR_ATmega328P__)
    // Arduino Duemilanove and Uno
    LINE("__AVR_ATmega328P__", __AVR_ATmega328P__);
  #endif
  
  #if defined(__AVR_ATmega2560__)
    // Arduino Mega 2560
    LINE("__AVR_ATmega2560__", __AVR_ATmega2560__);
  #endif
  
  #if defined(__AVR_ATmega1280__)
    // Arduino Mega
    LINE("__AVR_ATmega1280__", __AVR_ATmega1280__);
  #endif
  
  #if defined(__AVR_ATmega32U4__)
    // Arduino Leonardo
    LINE("__AVR_ATmega32U4__", __AVR_ATmega32U4__);
  #endif
  
  #if defined(__SAM3X8E__)
    // Arduino Due
    LINE("__SAM3X8E__", __SAM3X8E__);
  #endif
  
  // Determine board type
  #if defined(__AVR_ATmega168__) || defined(__AVR_ATmega328P__)
    Serial.println("Regular Arduino");
  #elif defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
    Serial.println("Arduino Mega"); 
  #elif defined(__AVR_ATmega32U4__)
    Serial.println("Arduino Leonardo");
  #elif defined(__SAM3X8E__)
    Serial.println("Arduino Due");
  #else
    Serial.println("Unknown");
  #endif
```

ianfixes:
from https://www.microchip.com/webdoc/avrlibcreferencemanual/using_tools_1using_avr_gcc_mach_opt.html

via @per1234 (https://github.com/ianfixes/arduino_ci/issues/86#issuecomment-454173396) 

Architecture | MCU name | Macro
-- | -- | --
avr1 | at90s1200 | __AVR_AT90S1200__
avr1 | attiny11 | __AVR_ATtiny11__
avr1 | attiny12 | __AVR_ATtiny12__
avr1 | attiny15 | __AVR_ATtiny15__
avr1 | attiny28 | __AVR_ATtiny28__
[...]

PR #123, opened 2019-05-29
Adding additional board definitions for #89

Doom4535:
## Highlights
Added entries for additional boards:
  * Nano (technically v3, v2.x looks to have used the atmega168)
  * Mega1280
  * Adafruit Grand Central M4 Express

## Issues Fixed
Changed processor used for Arduino due from atmega/avr to atmel sam

## Notes
* No action has been taken as of yet to expose additional analog pins on Nano
* No action has been take to expose additional features of the Grand Central beyond the standard M4 at this time

### DIFF
--- a/misc/default.yml
+++ b/misc/default.yml
@@ -92,6 +92,15 @@ platforms:
       defines:
       warnings:
       flags:
+  mega1280:
+    board: arduino:avr:mega:cpu=atmega1280
+    package: arduino:avr
+    gcc:
+      features:
+      defines:
+        - __AVR_ATmega1280__
+      warnings:
+      flags:
   mega2560:
     board: arduino:avr:mega:cpu=atmega2560
     package: arduino:avr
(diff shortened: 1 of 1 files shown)


Doom4535:
Welp, this is what I get for thinking these changes could be made from mobile only.

ianfixes:
This looks like a good start.  Don't be too concerned about the compiler errors, this simply means that we need to do some intelligent mocking out of `define`s, possibly in `cpp/arduino/ArduinoDefines.h`.  

The [most ominous warning](https://travis-ci.org/ianfixes/arduino_ci/jobs/538539802#L2996) (about  `device type not defined`) is coming direction from [`avr/io.h`](https://github.com/ianfixes/arduino_ci/blob/master/cpp/arduino/avr/io.h#L621-L625), where I don't see `__AVR_ATSAM3X8E__` listed.

Could it be that I'm using an outdated set of library files from the Arduino library (`io.h` in particular)?

ianfixes:
Reposting on #124 which seems to be a duplicate/repeat of this issue.

PR #124, opened 2019-05-29
Additional board definitions for #89

Doom4535:
## Highlights
Added entries for additional boards:
  * Nano (defaults to v3, includes v2 as well)
  * Mega1280
  * Adafruit Grand Central M4 Express

## Issues Fixed
Changed processor used for Arduino due from atmega/avr to atmel sam

## Notes
* No action has been taken as of yet to expose additional analog pins on Nano
* No action has been take to expose additional features of the Grand Central beyond the standard M4 at this time

### DIFF
--- a/misc/default.yml
+++ b/misc/default.yml
@@ -92,6 +92,15 @@ platforms:
       defines:
       warnings:
       flags:
+  mega1280:
+    board: arduino:avr:mega:cpu=atmega1280
+    package: arduino:avr
+    gcc:
+      features:
+      defines:
+        - __AVR_ATmega1280__
+      warnings:
+      flags:
   mega2560:
     board: arduino:avr:mega:cpu=atmega2560
     package: arduino:avr
--- a/misc/default.yml
+++ b/misc/default.yml
@@ -14,7 +14,16 @@ packages:
     url: https://dl.espressif.com/dl/package_esp32_index.json
 
 platforms:
-
+  
+  nano:
+    board: arduino:avr:nano
+    package: arduino:avr
+    gcc:
+      features:
+      defines:
+        - __AVR_ATmega328__
+      warnings:
+      flags:
   uno:
     board: arduino:avr:uno
     package: arduino:avr
@@ -30,7 +39,7 @@ platforms:
     gcc:
       features:
       defines:
-        - __AVR_ATmega328__
+        - __AVR_ATSAM3X8E__
       warnings:
       flags:
   zero:
@@ -86,6 +95,14 @@ platforms:
       flags:
   m4:
     board: adafruit:samd:adafruit_metro_m4
+    package: adafruit:samd
+    gcc:
+      features:
+      defines:
+      warnings:
+      flags:
+  grand_central_m4:
+    board: adafruit:samd:adafruit_grand_central_m4
     package: adafruit:samd
     gcc:
       features:
--- a/misc/default.yml
+++ b/misc/default.yml
@@ -15,7 +15,16 @@ packages:
 
 platforms:
   
-  nano:
+  nano_v2:
+    board: arduino:avr:nano
+    package: arduino:avr
+    gcc:
+      features:
+      defines:
+        - __AVR_ATmega168__
+      warnings:
+      flags:
+  nano_v3: &nano_p
     board: arduino:avr:nano
     package: arduino:avr
     gcc:
@@ -24,6 +33,7 @@ platforms:
         - __AVR_ATmega328__
       warnings:
       flags:
+  nano: *nano_p
   uno:
     board: arduino:avr:uno
     package: arduino:avr
@@ -39,7 +49,7 @@ platforms:
     gcc:
       features:
       defines:
-        - __AVR_ATSAM3X8E__
+        - __ATSAM3X8E__
       warnings:
       flags:
   zero:


Doom4535:
Issues seem to stem from changing the Due from the avr chip to the sam3x8
I believe the Arduino.h file used by Godmode for this will need to be swapped out for a ARM/SAM version from here: https://github.com/arduino/ArduinoCore-sam/blob/master/cores/arduino/Arduino.h

Although, the build seems to work fine for me on my local system.

ianfixes:
> Originally posted at https://github.com/ianfixes/arduino_ci/pull/123#issuecomment-496894658

This looks like a good start.  Don't be too concerned about the compiler errors, this simply means that we need to do some intelligent mocking out of `define`s, possibly in `cpp/arduino/ArduinoDefines.h`.  

The [most ominous warning](https://travis-ci.org/ianfixes/arduino_ci/jobs/538539802#L2996) (about  `device type not defined`) is coming direction from [`avr/io.h`](https://github.com/ianfixes/arduino_ci/blob/master/cpp/arduino/avr/io.h#L621-L625), where I don't see `__AVR_ATSAM3X8E__` listed.

Could it be that I'm using an outdated set of library files from the Arduino library (`io.h` in particular)?

Doom4535:
I believe that the `device type not defined` warning is related to the `ATSAM3X8E` not being an AVR board (none of the ARM boards are in `io.h)`.  I'm not entirely sure how the Arduino IDE does the build behind the scenes, but I want to say they switch to a separate ARM chain?

The Godmode is probably currently limited to AVR devices only?  How are the ESP builds handled?  I may try to use the same procedure for SAM devices.

.travis.yml
```diff
@@ -17,9 +17,13 @@ matrix:
     - os: osx
       env: BADGE=linux
 
-#before_install: gem install bundler -v 1.15.4
+before_install:
```
ianfixes: I'm not necessarily opposed to whatever is happening here but if you could give some context on how this relates to the problem being solved, that would be helpful (up to and including opening an issue).

ianfixes:
In this project (without official support from Arduino) I'm generating my best guess at the compiler commands from scratch.  I have to pull in a lot of the header files from the Arduino library to make this happen.  

This makes things a bit strange, because it means that I'm compiling & running the test on a desktop architecture -- not the Arduino target architecture(s).  So I need to bring in as many of the board-specific `#define`s as I can (from the official IDE's package), but not things like data types or other built-ins.

Nothing in Godmode itself should be limited to AVR devices.  Or rather, any functions that _are_ limited in such a way should be put behind the appropriate `#ifdef` directives, just like you would for (e.g.) serial port variable definitions.

Am I answering your questions?  I apologize, because faking out the compiler is the most complex and most fragile part of this project, despite all the work put into exposing the definitions via yaml config.

.travis.yml
Doom4535: I am indifferent to it ultimately, that is there because my CI is defaulting to bundler 2.01, which complains about not being compatible with the older version.  I set the version so that I don't cause any issues with anyone else.

Doom4535:
I believe so, the project is impressive.  Having an open source software-in-the-loop system is very exciting.

.travis.yml
ianfixes: If I'm understanding you correctly, there is a more recent version of bundler that we could be using?  I wouldn't mind taking some time to update that if it's causing problems -- please open an issue with the details of what you're seeing and I'll take a look at it.

ianfixes:
Thanks, that's nice of you to say 😄

lastthyme (approved):

ianfixes:
Picking up this project again after a hiatus.  @RobTillaart or @per1234 can you speak to whether the compiler flags listed here are comprehensive?  

@Doom4535 can you rebase this off what I'm working on in the #338 PR?

RobTillaart:
> Picking up this project again after a hiatus. @RobTillaart or @per1234 can you speak to whether the compiler flags listed here are comprehensive?

@ianfixes 
Not understand your question I'm afraid. 

If I need a board unique define (most often to make a library work)  I select that board in the IDE.
Then I compile the code - verbose mode always on.
I copy the output in my editor and search the -D flags to find "Arduino_ARCH_boardname". 
These defines allows me to add a patch for specific boards relative easily as these defines can be hard to find otherwise.
(or I just don't know where to look :)

Does this answer your question?

misc/default.yml
```diff
@@ -30,7 +49,13 @@ platforms:
     gcc:
       features:
       defines:
-        - __AVR_ATmega328__
+        # The due should use the ATSAM3X8E; however, that component does not seem
+        # to be defined for the Godmode tests/components
+        # The next closest part would be the AVR_ATmega2560 as both boards
+        # use the mega form factor
```
ianfixes, line 55: @Doom4535 can you explain this comment a bit?  I'm not sure why the fix belongs here and not in `Godmode.h`

Doom4535: It's been a while since I looked at this, I believe I made that change because the Arduino Duo is most like an Arduino Mega based on pinout. Updating [ `godmode.h`](https://github.com/Arduino-CI/arduino_ci/blob/29ab7b3949e95fb184b00b0809e8b53f1f0b2769/cpp/arduino/Godmode.h#L3-5) to include the equivalent for the SAM board would be the technically correct result, and allow for including support for any extra features the SAM board supports that can't be done with the AVR board.

I'm not sure what the replacement should be, on windows it would likely be something inside of `%LocalAppData%\Arduino15\packages\arduino\hardware\sam`, maybe `%LocalAppData%\Arduino15\packages\arduino\hardware\sam\1.6.12\variants\arduino_due_x\variant.h`?

PR #126, opened 2019-07-27
include ability to exclude certain directories from compilation

### LINKED ISSUE #29: Best practices to exclude folders from compilation?
## Issue / Feature Request Summary
A configuration parameter to exclude folders from compilation. 

Consider a case where all libraries are in `./lib` so they can all be in a single git hub repo (using git submodules). `arduino_ci` walks the tree to find all the header files - irrespective of any compiler directives to use mock headers/classes instead, the libraries get pulled in too (which may break compilation).

dolfandringa:
## Exclude certain directories from exclusion in compilation

* Add exclude_dirs option to [unittest] section in configuration file
* Add a filter in ArduinoCI::CppLibrary to filter out the files in directories specified there
* Adjust unittests to take the new feature into account

## Issues Fixed

* Fixes #29

### DIFF
--- a/SampleProjects/TestSomething/excludeThis/exclude-this.cpp
+++ b/SampleProjects/TestSomething/excludeThis/exclude-this.cpp
@@ -1,0 +1,10 @@
+#include "test-something.h"
+int testSomething(void) {
+  millis();  // this line is only here to test that we're able to refer to the builtins
+  return 4;
+};
+
+int* aNullPointer(void) {
+  int* ret = nullptr;
+  return ret;
+}
--- a/lib/arduino_ci/cpp_library.rb
+++ b/lib/arduino_ci/cpp_library.rb
@@ -37,11 +37,14 @@ module ArduinoCI
 
     # @param base_dir [Pathname] The path to the library being tested
     # @param arduino_lib_dir [Pathname] The path to the libraries directory
-    def initialize(base_dir, arduino_lib_dir)
+    def initialize(base_dir, arduino_lib_dir, exclude_dirs = [])
       raise ArgumentError, 'base_dir is not a Pathname' unless base_dir.is_a? Pathname
       raise ArgumentError, 'arduino_lib_dir is not a Pathname' unless arduino_lib_dir.is_a? Pathname
+      raise ArgumentError, 'exclude_dir is not an array of Pathnames' unless exclude_dirs.is_a?(Array) &&
+                                                                             exclude_dirs.each { |p| p.is_a? Pathname }
 
       @base_dir = base_dir
+      @exclude_dirs = exclude_dirs
       @arduino_lib_dir = arduino_lib_dir.expand_path
       @artifacts = []
       @last_err = ""
@@ -115,6 +118,19 @@ module ArduinoCI
       false
     end
 
+    # Guess whether a file is part of any @excludes_dir dir (indicating library compilation should ignore it).
+    #
+    # @param path [Pathname] The path to check
+    # @return [bool]
+    def in_exclude_dir?(path)
+      # we could do this but some rubies don't return an enumerator for ascend
+      # path.ascend.any? { |part| tests_dir_aliases.include?(part) }
+      path.ascend do |part|
+        return true if exclude_dir.any? { |p| p.realpath == part }
+      end
+      false
+    end
+
     # Check whether libasan (and by extension -fsanitizer=address) is supported
     #
     # This requires compilation of a sample program, and will be cached
@@ -150,7 +166,7 @@ module ArduinoCI
     # CPP files that are part of the project library under test
     # @return [Array<Pathname>]
     def cpp_files
-      cpp_files_in(@base_dir).reject { |p| vendor_bundle?(p) || in_tests_dir?(p) }
+      cpp_files_in(@base_dir).reject { |p| vendor_bundle?(p) || in_tests_dir?(p) || in_exclude_dir?(p) }
     end
 
     # CPP files that are part of the arduino mock library we're providing
@@ -172,6 +188,12 @@ module ArduinoCI
       arduino_library_src_dirs(aux_libraries).map { |d| cpp_files_in(d) }.flatten.uniq
     end
 
+    # Returns the Pathnames for all paths to exclude from testing and compilation
+    # @return [Array<Pathname>]
+    def exclude_dir
+      @exclude_dirs.map { |p| Pathname.new(@base_dir) + p }.select(&:exist?)
+    end
+
     # The directory where we expect to find unit test defintions provided by the user
     # @return [Pathname]
     def tests_dir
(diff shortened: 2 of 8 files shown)


ianfixes:
I want to sleep on this but I still want to help you with the problem you described in #29 .  Can we resolve that first (changing directories) and then resume here?

Immediate thoughts on this PR:

* Overall, I like it and I like your addition to the config file
* It looks like you're trying to add some directories (`excludeThis`) that, if included, would break the build and/or break the tests.  Some more explicit documentation around that would be helpful
* I'm nervous about the change to `CppLibrary`; I think that rather than array, this should probably accept some kind of block that selects/rejects files from the library.  But I have to think more about that.

dolfandringa:
@ianfixes I don't know what ticket you are looking at, but as far as I understand issue #29 is a feature request about excluding directories from the tree walk process when looking for CPP files. That is exactly what this ticket addresses. 

I modified the tests so they explicitly test my new feature. Without my feature, the tests would fail because "excludeThis" contains code that would make the compilation fail. 
Also in `spec/testsomething_unittests_spec.rb` the `cpp_files` test in 'TestSomething C++' has specifically been adjusted to test that the files in the 'excludeThis' folder are excluded from compilation. Without my feature, this test would now fail. The only thing I can think to change is to duplicate that test, once with, and once without the exclude_dirs parameter. Right now it's one combined test.

Documentation: I adjusted the README.md to document the feature. What else should I document?

lib/arduino_ci/cpp_library.rb
```diff
@@ -37,11 +37,14 @@ class CppLibrary
 
     # @param base_dir [Pathname] The path to the library being tested
     # @param arduino_lib_dir [Pathname] The path to the libraries directory
-    def initialize(base_dir, arduino_lib_dir)
+    def initialize(base_dir, arduino_lib_dir, exclude_dirs = [])
       raise ArgumentError, 'base_dir is not a Pathname' unless base_dir.is_a? Pathname
       raise ArgumentError, 'arduino_lib_dir is not a Pathname' unless arduino_lib_dir.is_a? Pathname
+      raise ArgumentError, 'exclude_dir is not an array of Pathnames' unless exclude_dirs.is_a?(Array) &&
+                                                                             exclude_dirs.each { |p| p.is_a? Pathname }
```
ianfixes: This should be [`.all?`](https://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-all-3F), not [`.each`](https://ruby-doc.org/core-2.4.1/Array.html#method-i-each).

SampleProjects/TestSomething/excludeThis/exclude-this.cpp
```diff
@@ -0,0 +1,10 @@
+#include "test-something.h"
+int testSomething(void) {
+  millis();  // this line is only here to test that we're able to refer to the builtins
+  return 4;
+};
+
+int* aNullPointer(void) {
+  int* ret = nullptr;
+  return ret;
+}
```
ianfixes: The contents of this directory appear to be a simple copy of files that already exist elsewhere in this project.

If the purpose of this file (and the `.h`) is just to cause some form of compilation error, I'd prefer that its contents speak entirely to that.  Just fill it with a syntax error in the form of 

```c++
This file was placed here specifically to break unit test compilation.  If arduino_ci is working properly, it should exclude this file (as per the .arduino-ci.yml configuration) and compilation should succeed. 
```

And name it something like `break-compilation.cpp`.

spec/testsomething_unittests_spec.rb
```diff
@@ -13,7 +13,9 @@ def get_relative_dir(sampleprojects_tests_dir)
 RSpec.describe "TestSomething C++" do
   next if skip_cpp_tests
   cpp_lib_path = sampleproj_path + "TestSomething"
-  cpp_library = ArduinoCI::CppLibrary.new(cpp_lib_path, Pathname.new("my_fake_arduino_lib_dir"))
+  cpp_library = ArduinoCI::CppLibrary.new(cpp_lib_path,
+                                          Pathname.new("my_fake_arduino_lib_dir"),
+                                          ["excludeThis"])
```
ianfixes: Please replace `["excludeThis"]` with `[]`, or add tests to the spec file that specifically make refer to the `excludeThis` directory and/or the files it contains.

lib/arduino_ci/cpp_library.rb
```diff
@@ -37,11 +37,14 @@ class CppLibrary
 
     # @param base_dir [Pathname] The path to the library being tested
     # @param arduino_lib_dir [Pathname] The path to the libraries directory
-    def initialize(base_dir, arduino_lib_dir)
+    def initialize(base_dir, arduino_lib_dir, exclude_dirs = [])
```
ianfixes: `exclude_dirs` should not be an optional argument here.  You'll need update all instances of `CppLibrary.new` (i.e. the spec files) to explicitly pass in the empty array.

ianfixes:
Thanks for contributing this, and I appreciate the attention you've paid to copying the existing style of the project.  I have a few small fixes to request, as your email inbox has probably told you already.

ianfixes:
Thanks for contributing this, and I appreciate the attention you've paid to copying the existing style of the project.  I have a few small fixes to request, and then this is good to merge.

spec/testsomething_unittests_spec.rb
ianfixes: To say this another way: I assume `["excludeThis"]` prevents a regression on the existing tests in this spec file.  That's fine, but you should add a different test section that shows that _without_ excluding a directory, `cpp_library.cpp_files` contains the extra files.

The easiest way I can think of would be to give each `context` section its own `cpp_library = ` line, instead of using a common one.

```ruby
  context "cpp_files without an exclude directory" do
    cpp_library = ArduinoCI::CppLibrary.new(cpp_lib_path, 
                                            Pathname.new("my_fake_arduino_lib_dir"),
                                            [])
    it "finds all cpp files in directory" do
      testsomething_cpp_files = [
        Pathname.new("TestSomething/test-something.cpp"),
        Pathname.new("TestSomething/excludeThis/break-compilation.cpp"),
      ]
      relative_paths = cpp_library.cpp_files.map { |f| get_relative_dir(f) }
      expect(relative_paths).to match_array(testsomething_cpp_files)
    end
  end

  context "cpp_files with an exclude directory" do
    cpp_library = ArduinoCI::CppLibrary.new(cpp_lib_path, 
                                            Pathname.new("my_fake_arduino_lib_dir"),
                                            ["excludeThis"])
    it "finds cpp files in directory and ignores some" do
      testsomething_cpp_files = [Pathname.new("TestSomething/test-something.cpp")]
      relative_paths = cpp_library.cpp_files.map { |f| get_relative_dir(f) }
      expect(relative_paths).to match_array(testsomething_cpp_files)
    end
  end

  context "unit tests" do
    cpp_library = ArduinoCI::CppLibrary.new(cpp_lib_path, 
                                            Pathname.new("my_fake_arduino_lib_dir"),
                                            ["excludeThis"])

```

ianfixes:
Merging to non-master branch as my requested edits are easy to make myself

PR #154, opened 2020-09-08
Provide for local Projects directory

jgfoster:
My reading of the docs leads me to believe that the typical usage is to create projects in `/SampleProjects/`. Doing so causes my checkout of the `arduino_ci` project to show changes. I suggest that we guide users to put their own projects in a `./Projects/` directory that is ignored by the `arduino_ci` project. To that end this PR adds that directory to the `.gitignore` file.

Also, I've found that `bundle install` doesn't work for me on macOS 10.15.6 since macOS tries to protect the global library. This PR also adds an alternate install script that places packages in the local directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ Gemfile.lock
 /coverage/
 /doc/
 /pkg/
+/Projects/
 /spec/reports/
 vendor
 *.gem
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ Windows  | [![Windows Build status](https://ci.appveyor.com/api/projects/status/
 
 ## Quick Start
 
+Create a project in (or copy your project to) the Projects directory. 
+
 For a bare-bones example that you can copy from, see [SampleProjects/DoSomething](SampleProjects/DoSomething).
 
 The complete set of C++ unit tests for the `arduino_ci` library itself are in the [SampleProjects/TestSomething](SampleProjects/TestSomething) project.  The [test files](SampleProjects/TestSomething/test/) are named after the type of feature being tested.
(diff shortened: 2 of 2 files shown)


ianfixes:
I think you are confusing the automated testing of _the source code for this ruby gem itself_ with how you'd test your own Arduino library _using_ this ruby gem.

The proper usage is to create libraries in your system's Arduino `libraries` directory.  The `arduino_ci` repository itself doesn't need to be on your system at all -- you need only a ruby environment and the metadata files (`Gemfile`, etc) to install and execute the test script.  `SampleProjects/DoSomething` is a _template_ for how this all works: it shows the metadata files files, expected directories, and sample CI configuration.

ianfixes:
Here are some examples of adding arduino_ci to existing projects:
* https://github.com/sdesalas/Arduino-Queue.h/pull/1/files
* https://github.com/adafruit/Adafruit_FONA/pull/91/files
* https://github.com/PaulStoffregen/OneWire/pull/67/files

jgfoster:
My experience (and expectation) with automated tests is that they are run on the developer's machine every few minutes, not just occasionally on some remote server following a git commit. Does `arduino_ci` support this approach? Is there any reason to discourage this approach?

ianfixes:
> automated tests [should] run on the developer's machine every few
minutes, not just occasionally on some remote server following a git
commit. Does arduino_ci support this approach?

Yes, you can run `bundle exec arduino_ci_remote.rb` locally.  Although I
had intended to have both a "local" and "remote" script, it turned out that
the "local" script had to do all the same checks as the remote one (I just
skip installation steps where appropriate).

jgfoster:
Yes, that makes sense now and I have a working case!

Issue #149, opened 2020-09-08
What is expected result of tests?


jgfoster:
## System

- OS: macOS Catalina (10.15.6)
- `ruby -v`: ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
- `bundle -v`: Bundler version 1.17.2
- `bundle info arduino_ci`: 
  * arduino_ci (0.3.0)
        Summary: Tools for building and unit testing Arduino libraries
        Homepage: http://github.com/ianfixes/arduino_ci
        Path: /Users/jfoster/code/arduino_ci
- `g++ -v`:
  Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
  Apple clang version 11.0.3 (clang-1103.0.32.62)
  Target: x86_64-apple-darwin19.6.0
  Thread model: posix
  InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
- Arduino IDE version: 1.8.13
- `java -version`:
  java version "11.0.7" 2020-04-14 LTS
  Java(TM) SE Runtime Environment 18.9 (build 11.0.7+8-LTS)
  Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.7+8-LTS, mixed mode)
- URL of failing Travis CI job: N/A
- URL of your Arduino project: https://github.com/Open-Acidification/Open_Acidification_pH-stat_arduino

## What is expected result of tests?

I'm looking at `arduino_ci` for our project and for a class I'm teaching this term. If we do use it I'll likely be making contributions and asking my students to do so as well (if nothing else, to documentation and samples). 

I've forked, cloned, and opened a terminal in SampleProjects/DoSomething. When I ran `bundle exec arduino_ci_remote.rb`, the result was a couple hundred lines of output, including the following:

```
...Unit testing bad-null.cpp with g++                                          ✗
...Unit testing bad-null.cpp with g++                                          ✗
...Unit testing bad-null.cpp with g++                                          ✗
WARNING: Spurious .bundle folder in 'DoSomething' library
WARNING: Spurious .bundle folder in 'TestSomething' library
[...]

Issue #152, opened 2020-09-08
Code style test finds 2160 offenses


jgfoster:
## System

- OS: macOS Catalina (10.15.6)
- `ruby -v`: ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
- `bundle -v`: Bundler version 1.17.2
- `bundle info arduino_ci`:
  arduino_ci (0.3.0)
  Summary: Tools for building and unit testing Arduino libraries
  Homepage: http://github.com/ianfixes/arduino_ci
  Path: /Users/jfoster/code/arduino_ci
- `g++ -v`:
  Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
  Apple clang version 11.0.3 (clang-1103.0.32.62)
  Target: x86_64-apple-darwin19.6.0
  Thread model: posix
  InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
- Arduino IDE version: 1.8.13
- `java -version`: 
  java version "11.0.7" 2020-04-14 LTS
  Java(TM) SE Runtime Environment 18.9 (build 11.0.7+8-LTS)
  Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.7+8-LTS, mixed mode)
- URL of failing Travis CI job: N/A
- URL of your Arduino project: https://github.com/Open-Acidification/Open_Acidification_pH-stat_arduino

## Numerous style test problems.

The [Contributing](https://github.com/ianfixes/arduino_ci/blob/master/CONTRIBUTING.md) guidelines say to run `bundle exec rubocop -D .` but when I do so after making a trivial change to one documentation file I get 2160 offenses. Does the CI job actually test this? Should they be fixed or should the contributing instructions be modified? Or is it something I'm doing wrong or misunderstand?

## Arduino or Unit Test Code, Illustrating the Problem

`bundle exec rubocop -D .`

ianfixes:
Rubocop is finding the vendor bundles in the sample projects, which wouldn't happen during CI.  In any case I'll merge #153 when it passes so that you don't experience this in the future.

PR #156, opened 2020-09-08
Update docs to explain purpose of SampleProjects

jgfoster:
This should avoid the impression that they are templates to be followed too closely. This is a partial fix for #149.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,10 +82,10 @@ vendor
 
 ### Installing the Dependencies
 
-Fulfilling the `arduino_ci` library dependency is as easy as running this command:
-
+Fulfilling the `arduino_ci` library dependency is as easy as running either of these two commands:
 ```
-$ bundle install
+$ bundle install   # adds packages to global library (may require admin rights)
+$ bundle install --path vendor/bundle   # adds packages to local library
 ```
 
 
--- a/SampleProjects/DoSomething/test/README.md
+++ b/SampleProjects/DoSomething/test/README.md
@@ -1,7 +1,9 @@
-# Purpose
+## Purpose
 
-These files are designed to test the Ruby gem itself, such that its basic tasks of library installation and compilation can be verified.  (i.e., use minimal C++ files -- feature tests for C++ unittest/arduino code belong in `../TestSomething/test/`).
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -82,10 +82,10 @@ vendor
 
 ### Installing the Dependencies
 
-Fulfilling the `arduino_ci` library dependency is as easy as running this command:
-
+Fulfilling the `arduino_ci` library dependency is as easy as running either of these two commands:
```
ianfixes, line 85: Conceptually, this is `bundle install`.  If you need an alternative that doesn't require admin access, `bundle install --vendor` is the second choice (aligning with #152) followed by `bundle install --path vendor/bundle/or/wherever`.

jgfoster:
That's fine. I'm happy to provide feedback in whatever manner seems easier for you. I appreciate the quick turn-around on the ideas and questions. I'm learning a lot, making good progress, and will have more submissions for your consideration soon.

README.md
jgfoster: On macOS 10.16, bundle version 1.17.2 gives `Unknown switches '--vendor'` in response to `bundle install --vendor`.  

As may be obvious, I'm not very familiar with Ruby, bundle, etc.

ianfixes: Sorry ... my head's in the wrong place. That should be `bundle install --deployment` 😅

README.md
```diff
@@ -82,10 +82,10 @@ vendor
...
 ```
-$ bundle install
+$ bundle install   # adds packages to global library (may require admin rights)
+$ bundle install --path vendor/bundle   # adds packages to local library
```
ianfixes, line 88: Did you ever attempt `bundle install --deployment` ?  I'm curious if that worked

jgfoster: Yes, and it did not work. I reported the same error as a simple install (asking for a password and suggesting the local path as an alternative).

jgfoster:
An unrelated matter (new issue?) but related to seeing things from the eyes of a new user:

I struggled with the compile vs. unit test options for a while. I spent a lot of time struggling with the compiler to get the unit tests to work and the idea that I'm running the unit tests without compiling (them?) seems very odd. Now that I'm understanding better I wonder if "Compile Examples" would be a better phrase (even though it is more verbose).

ianfixes:
Lets's absolutely open that new issue.  `SampleProjects` becomes `CompileExamples`?  I really like that idea.

PR #167, opened 2020-10-03
Introduce EEPROM support (#166)

jgfoster:
First, is this really as easy as it seems? I copied EEPROM.h from [here](https://github.com/arduino/ArduinoCore-megaavr/blob/c8a1dd996c783777ec46167cfd8ad3fd2e6df185/libraries/EEPROM/src/EEPROM.h), and added lines 30-35. It seems to compile and one function gives the expected value (obviously more testing needs to be done; see below). 

Is it possible to create a branch and merge this into the branch (unless you think it is okay in master)? I'd like to invite someone to add more tests and would prefer that they contribute here rather than through my fork (but that is a fine second-best alternative).

### DIFF
--- a/SampleProjects/TestSomething/test/eeprom.cpp
+++ b/SampleProjects/TestSomething/test/eeprom.cpp
@@ -1,0 +1,10 @@
+#include <ArduinoUnitTests.h>
+#include <Arduino.h>
+#include <EEPROM.h>
+
+unittest(length)
+{
+  assertEqual(EEPROM_SIZE, EEPROM.length());
+}
+
+unittest_main()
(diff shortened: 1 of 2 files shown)


ianfixes:
I wouldn't be surprised if this is how easy it was, it looks like EEPROM just provides an alternate storage mechanism.  I would defer to @per1234  on whether there are compiler directives that should block this code from being included on certain platforms.

Also, I'd prefer to add unit tests of (at least _some_) read, write, and comparison functions just to prove that the implementation -- obvious though it is -- matches what is expected.

jgfoster:
I agree that more tests are needed, but that is what students are for! I can have them contribute the additional tests to my private repository or to a branch here. I thought that a branch here would be more realistic. So would it be appropriate to have a branch that doesn't have adequate testing yet, and then once the branch is up to par it could be merged into master?

ianfixes:
Absolutely, in CONTRIBUTING.md I describe using a branch called `tdd` for that purpose.

jgfoster:
Thanks. I see now how to submit a PR to a branch. I'll do that...

ianfixes:
Just for git's issue-linking visibility, this was to address #166

Issue #170, opened 2020-10-05
Conflicting definitions of `ostream& operator<<()` with Xcode 12.0.1


jgfoster:
## System

- OS: macOS 10.15.7
- `ruby -v`: ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
- `bundle -v`: Bundler version 1.17.2
- `bundle info arduino_ci`: arduino_ci (0.3.0)
- `g++ -v`: Apple clang version 12.0.0 (clang-1200.0.32.2)
- Arduino IDE version: 1.8.13
- `java -version`: java version "11.0.7" 2020-04-14 LTS
- URL of failing Travis CI job: (none)
- URL of your Arduino project: https://github.com/jgfoster/arduino_ci

## Conflicting definitions of `ostream& operator<<()`

The following both have definitions:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/ostream:223:20
/Users/jfoster/code/Arduino/arduino_ci/cpp/unittest/OstreamHelpers.h:5:22
```

Here is the output:

```
Unit testing library.cpp with g++... 

[...]

PR #171, opened 2020-10-05
Don't define ostream& operator() if already defined by Apple

jgfoster:
Possible fix for #170. I'm not sure when Xcode started providing this, but it feels recent and this change allows tests to pass for me on my local build. I've tried to make it safe for everyone else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Removed
 
 ### Fixed
+- Don't define `ostream& operator<<(nullptr_t)` if already defined by Apple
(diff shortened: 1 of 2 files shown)


ianfixes:
This seems like the right approach for now, can you just note it in CHANGELOG.md and then I'll merge?

ianfixes:
Are you blocked until a new gem is released with this fix?

jgfoster:
I don't think so. I haven't seen this in my external library development, only when I'm testing `arduino_ci` itself (through `SampleProjects/TestSomething`). So I can continue my separate development in the old gem and my internal development in the (newly-merged) master. At least that is the way it seems to me at the moment!

PR #172, opened 2020-10-05
Bring branch `tdd` up to master

jgfoster:
Note that the `tdd` branch is far behind master and this will bring it current. Also, note that the branch should be used for only one "broken" feature at a time (else there will be much confusion and it will be difficult to merge to master from this branch).


ianfixes:
Accomplished via force pushing `master` to `tdd`

PR #177, opened 2020-10-12
Add tests for EEPROM

Lizj96:
tests the public API

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,10 +82,10 @@ vendor
 
 ### Installing the Dependencies
 
-Fulfilling the `arduino_ci` library dependency is as easy as running this command:
-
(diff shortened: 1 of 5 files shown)


jgfoster (approved):
@ianfixes, With this PR we have tests of the six public APIs for EEPROM and we try read/write of various values and at various points in memory. From my understanding of what EEPROM is supposed to do, we have it implemented and tested completely. Unless you have further suggestions (which are welcome!), I believe this should be merged into `tdd` and then into `master` (I can submit a new PR for that after this is merged if you like). Thanks!

ianfixes:
This looks great, I see the passing tests!

ianfixes:
Sorry for the delay on this

PR #183, opened 2020-10-16
Suggested fixes

ianfixes:
### Trying it out
You can test this beta of the gem directly from github by putting this in your `Gemfile`:

```ruby
gem 'arduino_ci', git: 'https://github.com/ianfixes/arduino_ci.git', branch: '2020-10-16_suggestions'
```

### Added
- `arduino_ci_remote.rb` CLI switch `--skip-examples-compilation`
- `CppLibrary.header_files` to find header files
- `LibraryProperties` to read metadata from Arduino libraries
- `CppLibrary.library_properties_path`, `CppLibrary.library_properties?`, `CppLibrary.library_properties` to expose library properties of a Cpp library
- `CppLibrary.arduino_library_dependencies` to list the dependent libraries specified by the library.properties file

### Changed
- `CppLibrary` functions returning C++ header or code files now respect the 1.0/1.5 library specification

### Deprecated
- `arduino_ci_remote.rb` CLI switch `--skip-compilation`

### Fixed
- Don't define `ostream& operator<<(nullptr_t)` if already defined by Apple
- `CppLibrary.in_tests_dir?` no longer produces an error if there is no tests directory
- The definition of the `_SFR_IO8` macro no longer produces errors about rvalues

* See CHANGELOG.md for more

## Issues Fixed

* Fixes #175
* Fixes #168
* Fixes #187
* Fixes #196 
* Fixes #140
* Fixes #166

PR #184, opened 2020-10-16
Add support for EEPROM

### LINKED ISSUE #166: Support for EEPROM

jgfoster:
Fixes #166. Also add testing to demonstrate use.


ianfixes:
What information should I put into `REFERENCE.md` for this?  It looks like it mocks the functionality, but does not interact with `GODMODE` in a way that would let you verify that the functions had been called a certain way.  (which does not block this PR)

jgfoster:
Actually it is fair for you to say that we need to update the documentation! How that looks is open for discussion and advice. From the beginning (only a few weeks ago) I've been unsure on how to fit in new mocks. Since the EEPROM functionality doesn't involve pins, I don't see the direct tie-in to `GODMODE`. As to verifying "that the functions had been called in a certain way," I'm not sure. In this case, the API is pretty much read/write. If there data is there, then then I assume it was called correctly. What more can be tested? (Not intended as pushing back, just trying to figure it out myself.)

Shall I put in a PR with a paragraph describing how to use EEPROM?

ianfixes:
Apologies for the delay.  It looks like the unit tests are all there, so all that I think remains is a short paragraph in `REFERENCE.md` similar to the section on SPI -- just saying that it acts like a basic read/write buffer and there is no GODMODE support at this time.  (That sort of thing can always be added later -- we can open an issue to track it if/when that functionality is desired).

jgfoster:
Add comments (#190).

ianfixes:
I'll smooth over the merge conflicts here in a bit

cpp/arduino/EEPROM.h
```diff
@@ -0,0 +1,166 @@
...
+  Original Copyright (c) 2006 David A. Mellis.  All right reserved.
+  New version by Christopher Andrews 2015.
+
+  Copy of https://github.com/arduino/ArduinoCore-megaavr/blob/c8a1dd996c783777ec46167cfd8ad3fd2e6df185/libraries/EEPROM/src/EEPROM.h
+  modified by James Foster in 2020 to work with Arduino CI.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
```
jgfoster, line 12: We should probably reconsider adding this directly since its license is inconsistent.

ianfixes: I'm really out of my depth with regard to the terms of all the different licenses, and really can't offer advice one way or another.  I have no idea whether your own rewrite counts as some "derivative" in the way the licenses are written.

I think as long as we're doing our best and intend to respond to any legal complaints -- fixing the license as appropriate -- (which I do intend to do) then I'm not sure what else it would take to move forward.  In any case, if we keep the file let's keep the license and it can all get sorted out later.

jgfoster: As owner of the project, you certainly have the last word on this. But (did you sense a "but" coming?), the difference between the GNU license and your Apache license is night and day (with _slight_ exaggeration, think Catholics and Protestants in 16th, 17th, and 18th century Europe, or Antifa vs. Proud Boys today). While I can't promise that my replacement would withstand a legal challenge, I'm pretty confident that it would be better to leave out the GNU reference. Other than the file name and the published API, I made an effort to write the code from scratch and think that the risk is extremely small. By leaving in the GNU license, you risk attaching GNU to my code, and then you would need another rewrite.

ianfixes: I saw your rewrite after adding my comment, and I think your rewrite satisfies

ianfixes:
I've squashed this into #183

Issue #133, opened 2019-08-08
Improve mocks of builtin macros


ianfixes:
A post to the arduino developers' mailing list suggests a smarter way to do macros for (e.g.) `max` using "statement expressions":

```c++
#define max(a,b) \
  ({ typeof (a) _a = (a); \
      typeof (b) _b = (b); \
    _a > _b ? _a : _b; })
```

Adopting this will prevent double-evaluating `a` and `b` if for some reason they themselves are expressions with side effects.

PR #189, opened 2020-10-26
High-level wire implementation, complete with unit tests

### LINKED ISSUE #181: Implement Wire
`cpp/arduino/Wire.h` has a number of `// TODO` comments. I'd like to complete the implementation and would appreciate guidance (or at least feedback on some ideas).

One extreme would be to pull in an existing implementation with the assumption that a "real" implementation would work by reading and writing pins (pins 20 & 21 on the Duo and Mega2560). With that, it is just like any other code that reads and writes pins.

A middle ground would be to write a new implementation that reads and write pins, but that seems like duplicate work.

The other extreme would be to mock the API at a higher level (and not set pins). Tests would not look at pins but would need to call a new testing API that could get and set values at a higher level than bits on a pin. This could be something added to Godmode or to Wire itself. 

I propose a new high-level API (I would want it anyway) and that it be in Wire (I'd prefer not to further pollute Godmode). Thoughts?

lsaca05:
Fixes #181 

Tracks addresses read from/written to on a high level using maps and vectors, as well as the content at that address, of course.

Let me know if you'd like more tests, though I think I covered everything that we deemed useful at the moment.


lsaca05:
I'm guessing the test failures are related to me overwriting the "map" from AvrMath with the class template from C++ (lines 25-27). I'm assuming there isn't as shorthand a way to fix it like the "max" function in lines 7-22.

jgfoster:
Rather than removing functionality that is part of the core Arduino system ([map](https://github.com/arduino/reference-en/blob/master/Language/Functions/Math/map.adoc)), we should use the C++ namespace feature to avoid the name conflict. So, remove line 36 and replace all your references to `map` with `std::map`. Then remove lines 24-29.

lsaca05:
> Rather than removing functionality that is part of the core Arduino system ([map](https://github.com/arduino/reference-en/blob/master/Language/Functions/Math/map.adoc)), we should use the C++ namespace feature to avoid the name conflict. So, remove line 36 and replace all your references to `map` with `std::map`. Then remove lines 24-29.

I tried this and my terminal gets flooded with errors related to the Arduino core map being used incorrectly. I deemed that ifdef patch job necessary at first, but I'll try and play around with namespaces to confirm when I get a chance.

jgfoster:
Ah, yes, I see. The existing macro will indiscriminately changes to all your uses of `map` and that's bad. And, you are right, the best solution would be something like what was done with `max()`. Absent that, we may need to come up with something else. This requires more thought.

jgfoster:
Let's try to rewrite this to use something other than a `map`.

SampleProjects/TestSomething/test/wire.cpp
```diff
@@ -0,0 +1,68 @@
+#include <ArduinoUnitTests.h>
+#include <Arduino.h>
+#include <Wire.h>
+using namespace std;
```
jgfoster, line 4: Let's remove this line (`using namespace std;`) and add `std::` to the front of each reference to `deque`. (This isn't very important but I'd like an edit to add a newline to the end of the file and sometimes git doesn't recognize that as enough of a change to commit/push).

SampleProjects/TestSomething/test/wire.cpp
```diff
@@ -0,0 +1,68 @@
...
+    assertEqual(2, mosi->size());
+    assertEqual(4, mosi->front());
+    mosi->pop_front();
+    assertEqual(1, mosi->size());
+    assertEqual(6, mosi->front());
+    mosi->pop_front();
+    assertEqual(0, mosi->size());
+}
+
+unittest_main()
```
jgfoster, line 68: And a newline at the end of the file.

jgfoster:
@ianfixes, With a couple minor edits, this is an implementation of Wire that I'd like to have you consider. Please let me know your thoughts. Thanks!

cpp/arduino/Wire.h
```diff
@@ -1,138 +1,265 @@
...
+ * structures, each of which contains a deque for input and a
+ * deque for output. You can preload data to be read and you can
+ * look at a log of data that has been written.
+ */
 
 #pragma once
 
 #include <inttypes.h>
+
+// https://github.com/Arduino-CI/arduino_ci/issues/165
```
ianfixes, line 38: I'm curious whether the `max` and `min` definitions here should replace my current definitions in the library itself -- whether they would pass all the existing tests.  I'm thinking of using this to resolve #133 (i.e. spreading this fix more broadly).

jgfoster: I would very much like to have the fix spread more broadly (we are having to include it in several libraries)! I got the code from #144 (which might be languishing because it is too large!?).

ianfixes: I hadn't considered merging #144 because the CI was failing -- I assumed it was still a work in progress.  I'll follow up on that, because I very much support the direction being taken.

jgfoster: And now that I look more carefully, I see the most recent commit comment is "Debug commit - do not merge"!

cpp/arduino/Wire.h
```diff
@@ -1,138 +1,265 @@
...
+#else
+#define min(a, b)                                                              \
+  ({                                                                           \
+    __typeof__(a) _a = (a);                                                    \
+    __typeof__(b) _b = (b);                                                    \
+    _a < _b ? _a : _b;                                                         \
+  })
+#endif
+#endif
+
```
jgfoster, line 72: @lsaca05, Could you merge the latest `master` branch into your `wire` branch and then see if you can remove these lines (they should now be unnecessary). Thanks!

jgfoster:
Looks good. Thanks!

cpp/arduino/Wire.h
```diff
@@ -1,138 +1,228 @@
...
+    return requestFrom(address, quantity, true);
   }
   uint8_t requestFrom(int address, int quantity) {
-    int stop = true;
-    return requestFrom(address, quantity, stop);
-  }
-  uint8_t requestFrom(uint8_t address, uint8_t quantity) {
-    return requestFrom((int)address, (int)quantity);
+    return requestFrom((uint8_t)address, (size_t)quantity);
   }
```
jgfoster, line 139: While this is a legitimate function "overload" in that `size_t` and `int` are different types, I think that the C++ compiler ought to be able to make implicit conversions for us. Furthermore, and more importantly, if we don't provide all possible alternatives to the integer types, then there will be an ambiguity error. I found this when someone calls this function and passes in a byte. I think it would be best to just have one function and let the C++ compiler do the type cases. Thanks!

lsaca05: Which function should remain?

SampleProjects/TestSomething/test/wire.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <Arduino.h>
+#include <Wire.h>
+using std::deque;
+
+unittest(begin_write_end) {
+    deque<uint8_t>* mosi = Wire.getMosi(14);
+    assertEqual(0, mosi->size());
+    Wire.begin();
+    Wire.beginTransmission(14);
+    Wire.write(0x07);
```
ianfixes, line 11: There are a lot of random-looking numbers in this unit test, which might be better expressed as named variables.  Otherwise, please add some comments that explain the different sections of the tests

ianfixes:
I appreciate the work that has gone into this, especially the unit tests that cover the added functionality.  @jgfoster let me know when your comments are addressed

cpp/arduino/Wire.h
jgfoster: Sorry for the ambiguity! The function with the smaller types calls the function with the larger types. Let's remove the one with the smaller types (`uint8_t`).

SampleProjects/TestSomething/test/wire.cpp
jgfoster: Rather than named variables, let's just add comments along the line of "try some random value (0x07) sent to some random slave (14)."

jgfoster:
@ianfixes, if you are satisfied with the comments in the test file, then I'm happy to have it merged. Thanks!

SampleProjects/TestSomething/test/wire.cpp
ianfixes: With the aid of your comments I can see what you're after.  But I'm going to insist on using named constants. 
 
The reason for this is purely readability; as a maintainer I have to scan a lot of contributions in a short amount of time and solve problems in them.  Consider the following chunk of this test as currently written:

```c++
    assertEqual(1, Wire.read());
    assertEqual(2, Wire.available());
    assertEqual(4, Wire.read());
    assertEqual(1, Wire.available());
    assertEqual(7, Wire.read());
```

This requires me (and any future contributors to this code) to puzzle out things like 
* the `1`, `4`, and `7`, here being conceptually related to each other (and to a definition a dozen lines earlier) _even though the values no longer appear on adjacent lines_ 
*  the `1` in `assertEqual(1, Wire.read());` and `assertEqual(1, Wire.available());` represent completely different things _even though they are are expressed in the exact same way_

This affects numeric tests more than strings or objects (which end up being a bit more self-descriptive).

Here's an example of how I'd like this to be written:

```c++
    const uint8_t randomSlave = 14;
    const uint8_t randomValues[] = { 0x07, 0x0E };
    Wire.begin();
    Wire.beginTransmission(randomSlave);
    Wire.write(randomValues[0]);
    Wire.write(randomValues[1]);
    Wire.endTransmission();

    // check master write buffer values
    assertEqual(2, mosi->size());
    assertEqual(randomValues[0], mosi->front());
    mosi->pop_front();
    assertEqual(randomValues[1], mosi->front());
    mosi->pop_front();
    assertEqual(0, mosi->size());
```

Similar for the values `19` and `34` -- these should be named variables, so that the function calls can be more indicative of what is being input.

```c++
    Wire.begin();
    deque<uint8_t>* miso;

    // place some values on random slaves' read buffers
    const int aRandomSlave = 19;
    const int anotherRandomSlave = 34;
    const uint8_t moreRandomValues = { 1, 4, 7 };
    miso = Wire.getMiso(aRandomSlave);
    miso->push_back(randomValues[0]);
    miso->push_back(randomValues[1]);
    miso = Wire.getMiso(anotherRandomSlave);
    miso->push_back(moreRandomValues[0]);
    miso->push_back(moreRandomValues[1]);
    miso->push_back(moreRandomValues[2]);
```

I chose `random` in the naming to reflect your comment, but `expected` (e.g. `expectedIntValues` / `expectedByteValues`) or `input` also work.  Or for things where you just have to pick a wire or port to work with, `const int arbitraryIndex = 19`, etc.  What needs to be conveyed is the difference between a specific value (e.g. a size) that's expected to be returned and an arbitrary number that just needs to match the input.

ianfixes:
Thanks for taking the time to add clarity to those test cases!

Issue #165, opened 2020-09-29
max() macro does not play well with std::max() 


jgfoster:
If I include `<Arduino.h>` before `<vector>`, then I get an [error](https://github.com/jgfoster/LiquidCrystal/runs/1180401006?check_suite_focus=true) in `stl_vector.h:1645` where the code refers to std::max().

```
	const size_type __len = size() + std::max(size(), __n);
```

This appears to be due to a macro definition of `max()` in `arduino_ci-0.3.0/cpp/arduino/AvrMath.h` causing the code to be expanded to:

```
	const size_type __len = size() + std::((size())>(__n)?(size()):(__n));
```

This produces `error: expected unqualified-id before ‘(’ token` (which makes sense).

My [research](https://stackoverflow.com/questions/51641994/error-defining-min-max-macro-in-f2c-h-file) suggests that this sort of use of macros should be avoided.

```
Unit testing BaseClass.cpp with g++... 

[...]

PR #194, opened 2020-11-05
Prevent name clashes with math macros

### LINKED ISSUE #133: Improve mocks of builtin macros

jgfoster:
This takes the edits from [here](https://github.com/matthijskooijman/arduino_ci/commit/e356ad78f36720d66ff46551d444f1a3687a484b) and makes them a separate PR.

Fixes #133 (and #165), simplifies #189, and replaces part of #144.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Move repository from https://github.com/ianfixes/arduino_ci to https://github.com/Arduino-CI/arduino_ci
+- Revise math macros to avoid name clashes
 
 ### Deprecated
 
--- a/cpp/arduino/AvrMath.h
+++ b/cpp/arduino/AvrMath.h
@@ -1,26 +1,126 @@
 #pragma once
+#include "ArduinoDefines.h"
 #include <math.h>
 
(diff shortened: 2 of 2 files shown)


ianfixes:
I see both a template and a macro definition of `max`.  If both are needed, can you add some code comments explaining that?

ianfixes:
Oh, I see what's happening: `#ifdef __cplusplus` wraps the whole section.  Can you indent the "body" of both the `#ifdef` and the `#else` sections of `__cplusplus` to make them more evident?

jgfoster:
Indent added (note that it would be undone by `clang-format`).

ianfixes:
I know the formatting isn't standard.  There is a lot of c++-side stuff that I'd prefer to keep on the readability side, since this project is still working its way to feature-completeness.  Once it works, I'm hoping the larger group of contributors that have joined in will be able to help me take a second look at how it all could be made more solid.

The end goal would be some direct tie-in with the Arduino project itself to enable these sorts of tests in a batteries-included sort of way in the IDE.  But that's a long way off, if ever.

PR #195, opened 2020-11-05
Rename `arduino_ci_remote.rb` to `arduino_ci.rb`

jgfoster:
One script to rule them all!


exe/arduino_ci_remote.rb
```diff
@@ -1,394 +1,3 @@
...
-config = ArduinoCI::CIConfig.default.from_project_library
-
-@arduino_cmd = ArduinoCI::ArduinoInstallation.autolocate!
-inform("Located Arduino binary") { @arduino_cmd.binary_path.to_s }
-
-perform_unit_tests(config)
-perform_compilation_tests(config)
-
-terminate(true)
+puts "arduino_ci.rb is deprecated in favor of arduino_ci.rb."
```
ianfixes, line 2: one typo

jgfoster: Good catch!

jgfoster:
It occurs to me that I should not have clicked "Resolved" on the review. I intended to communicate that I believe I had addressed the issue raised!

Issue #187, opened 2020-10-26
Fix _SFR_IO8 macro to avoid "cannot take the address of an rvalue"


ianfixes:
## Issue Summary 

see https://github.com/Arduino-CI/SD/issues/2

```
error: cannot take the address of an rvalue of type 'int'
  {&DDRE, &PINE, &PORTE, 0},  // E0  0
   ^~~~~
```

[Working theory](https://github.com/Arduino-CI/SD/issues/2#issuecomment-716266934) is that this is caused by a badly formed macro. 

https://github.com/Arduino-CI/arduino_ci/blob/master/cpp/arduino/avr/io.h#L99

```diff
- #define _SFR_IO8(io_addr) (io_addr) // this macro is all we need from the sfr file 
+ #define _MMIO_BYTE(mem_addr) (*(volatile uint8_t *)(mem_addr))
+ #define _SFR_IO8(io_addr) _MMIO_BYTE((io_addr) + __SFR_OFFSET)
```

Will see if `volatile` is sufficient to fix

jgfoster:
Also see #120. I copied code from there and got things to compile.

ianfixes:
I was able to reproduce this with a unit test, and added the `volatile` keyword (and cast) to fix the error.  Currently in my feature branch.

ianfixes:
This took a lot longer than expected due to chasing my tail on linker errors (and segfaults).  I think I am out of the woods now though.

PR #201, opened 2020-11-09
Possible direction for `__ARDUINO_CI_SFR_MOCK`

jgfoster:
This seems to hint at a direction to take for #187. Before merging please see `SPI.h:100` where a change was made in order to get the `spi` test in `godmode.cpp` to pass. It is almost certainly wrong, but I don't know enough to fix it. 

This is also related to discussions in #183.


matthijskooijman:
For the SPI thing, see the related commit https://github.com/Arduino-CI/arduino_ci/pull/144/commits/90dc7476111ca90f248836a68597a3d18360ac91 from #144.

I suspect that:
 - With the old definition, any IO register would resolve to its address, rather than a meaningful value.
 - For the 328p, this means that `SPCR` would resolve to 0x2C: https://github.com/Arduino-CI/arduino_ci/blob/f6ca7a76886aa03bc1c1880ea48503ae14822ea4/cpp/arduino/avr/iom328p.h#L287
 - Since `1 << DODR == 0x20`, the SPI library check `if (!(SPCR & (1 << DORD))) {` would end up as `false`, ending up in the `else`.
 - With this PR, `SPCR` is actually being read, but since nothing writes it, this ends up reading `0`, and the `if` ends up as `true`, using the wrong byte order.
 - This `if` is in the mock SPI library to support sketches that call `SPI.begin()` and then manually change `SPCR` to use a non-standard byte order (lacking a standard Arduino API to do this). Note that with the current `_SFR_IO8`, those sketches could never have worked, though, since assigning `SPCR` breaks.

If the above is correct, I think a correct fix could be to set a meaningful value for `SPCR` in the `SPI.begin()` method. At least `DODR` should be set, maybe the (code to calculate the) value should just be copied from the AVR SPI library? This only applies to `__AVR__`, since the `SPCR` "API" is AVR-specific (so both writes and reads from `SPCR` should be guarded by `__AVR__`, like my commit does for the read).

ianfixes:
If we rebase this off master (which should now have the `__AVR__` guard in place) will that resolve all the concerns here?

jgfoster:
I've taken a bit of a different approach than suggested by @matthijskooijman above. Instead of using `(SPCR & (1 << DORD)` to determine `MSBFIRST` or `LSBFIRST` I am saving it from the `beginTransmission()` call (the previous mock code was simply ignoring the parameters). This makes more sense to me but that doesn't mean that it is right! 

Another area that should be considered is the `long long` at the end of `Godmode.cpp`. There was a suggestion that it be `uint8_t` (bytes) since this more accurately represents the actual registers. On the other hand, I notice that there are attempts to use 16-bit or 32-bit pointers and that could cause an alignment error.

But, from what I have, I'm satisfied with this PR and I think it is an improvement on the previous situation.

Advice, as always, is welcome.

ianfixes:
> I think it is an improvement on the previous situation

As do I... my round of edits on this all ended in some segfaulting unit tests.  I've squashed this into #183 and we'll see where it takes us

cpp/arduino/SPI.h
```diff
@@ -94,10 +98,7 @@ class SPIClass: public ObservableDataStream {
   uint16_t transfer16(uint16_t data) {
     union { uint16_t val; struct { uint8_t lsb; uint8_t msb; }; } in, out;
     in.val = data;
-    // Changes in the definition of _SFR_IO8() cause this to break.
-    // This horible hack allows the tests to pass but is done 
-    // without any understanding at all of what is going on here!
-    if (false && !(SPCR & (1 << DORD))) {
+    if (bitOrder == MSBFIRST) {
```
matthijskooijman, line 101: I don't think this is correct. AFAIU, `MSBFIRST` / `LSBFIRST` are about the  bit order, so the order of bits within a byte. This if is about byte order, so the order of bytes within a 16-bit word. AFAIK, the official Arduino API lacks a way to configure this byte order, but on AVR the hardware handles this, so sketches and libraries can mess with the `SPCR` register to let the hardware reverse the bit order.

matthijskooijman:
> Another area that should be considered is the long long at the end of Godmode.cpp. There was a suggestion that it be uint8_t (bytes) since this more accurately represents the actual registers. On the other hand, I notice that there are attempts to use 16-bit or 32-bit pointers and that could cause an alignment error.

Good point about alignment. Maybe the AVR registers happen to be aligned, but I wouldn't really expect so (and if so, probably only 16-bit, not 32-bit). I was thinking reading registers could be more explicit, e.g.:

```

 #define _SFR_IO16(io_addr) ((uint16_t)__ARDUINO_CI_SFR_MOCK[io_addr + 1] << 8 | __ARDUINO_CI_SFR_MOCK[io_addr]))
```

But that no longer allows assigning values to the register, so that won't fly...

I'm not actually sure how to handle this in a way that:
 - Allows assignment
 - Does not violate alignment
 - Merges two adjacent `_SFR_IO8` accesses in a single `_SFR_IO16` access.
 - Does not violate strict aliasing rules (can probably be done using a union)
 - Works in C (in C++, you can probably pull this off with custom assignment operators)

Mostly alignment is a problem, anyone know what the rules are for that? IIRC on some platforms misaligned access is a fatal problem, on a lot of others it's just slower. Can you somehow tell the compiler that you're not sure a pointer is aligned so it can generate slower but safe byte-by-byte accesses automatically maybe?

jgfoster:
It appears that x86 allows misaligned memory accesses [stackoverflow](https://stackoverflow.com/questions/1237963/alignment-along-4-byte-boundaries), so I think that changing this to a byte would be safe and accurate. I'll update this PR.

cpp/arduino/SPI.h
jgfoster: You are right, the flags are for bits not bytes. I'll follow your earlier suggestion of setting and checking the `SPCR` register and try to build some tests. Also, that will require following your (even) earlier code of checking for `SPCR` and `DORD` being defined. It will be a bit more complicated, but I'll give it a go!

jgfoster: Further investigation suggests that `DORD` is the _bit order_, not _byte order_ (see [here](http://avrbeginners.net/architecture/spi/spi.html) and [here](https://ww1.microchip.com/downloads/en/AppNotes/Atmel-2585-Setup-and-Use-of-the-SPI_ApplicationNote_AVR151.pdf)). So, I'm again confused. 

Maybe if you are doing most significant bit first then you need to get the most significant byte first and if you are doing the least significant bit first then you need to start with the least significant byte. In that case the code is okay? But it could be improved by using `SPCR` and `DORD`.

jgfoster:
@matthijskooijman, This time I've taken the approach you suggested above (I appreciate your patience and explanations). Further review and comments are welcome.

@ianfixes, I've added tests for bit/byte order and for misaligned memory accesses. I think this a further improvement and invite you to merge it with the previous code.

cpp/arduino/SPI.h
matthijskooijman: Right, maybe I was working too much off assumptions based on the mocked SPI.h :-)

So then I guess `DORD` actually decides the bit order, which also implies a particular the byte order. And since the mocked SPI interface is not a physical one where bit order plays a role, the bit order is actually not so relevant (since we just store stransmitted bytes in memory), but the byte order is (since we split a 16-bit word into two bytes).

Looking at the actual SPI implementation (https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/SPI/src/SPI.h) it seems that, unlike what I assumed, the hardware does not actually implement 16-bit transfers, this is done in software. So the official AVR `transfer16` also checks `DORD` to see whether it should send the MSByte or LSByte first (so basically it just lets the byte order follow the configured bit order).

Then, what I said before about there not being an API to set the byte order, so sketches would write `DORD` also seems to be untrue: The byte order always follows the bit order, so the byte order can be set by setting the bit order. Even more, the AVR SPI code checks `DORD` in `transfer16`, but `DORD` is set in e.g. `beginTransaction` based on `MSBFIRST` / `LSBFIRST`. IOW, they just check `DORD` in `transfer16` so they don't need to store `MSBFIRST` / `LSBFIRST` anywhere.


Summarizing: I was quite wrong about nearly everything, and your [previous fix that remembered `bitOrder`](https://github.com/Arduino-CI/arduino_ci/pull/201/commits/57070abdd6cdeca7ac68c391dccd68aa5963b1c5) actually seems like the correct one (so I would suggest reverting the SPI-related changes in your most recent commit).

An unrelated note: Your last commit contains two fairly separate changes, it would be easier to review if they were in separate commits (in git, the first line is considered the title and is the only one shown in a lot of places, so I initially thought that you maybe did not push your SPI changes yet). Also, if these commits are merged as-is, having such a double commit in the official history makes it harder to look through the commit history later (I'd personally restructure them, squashing multiple revisions of the same fix and separating out different fixes before merging, though).

jgfoster: I've reverted the `bitOrder` code and will keep in mind the commit history suggestions. Thanks for the feedback!

ianfixes:
I think I have this all picked in again

ianfixes:
It's failing here:
https://ci.appveyor.com/project/ianfixes/arduino-ci/builds/36287425#L3081

I'm not sure whether that's because a separate patch in `SPI.h` had me wrapping the `bitOrder == MSBFIRST` test in a preprocessor check:

```c++
    #if defined(SPCR) && defined(DORD)
    if (bitOrder == MSBFIRST) {
```

Is ` defined(SPCR) && defined(DORD)` still relevant here?  I'm going to try without anyway, but I'm curious if that's correct.

jgfoster:
Yes, the failing tests are related to confirming that the two bit-order configs each work. When the code used `SPCR` and `DORD` then they needed to be guarded with a `defined()` check. But now that we are back to using a private instance variable to remember the bit order, there is no need to guard the bit order check.

matthijskooijman:
Agreed, the changes from this PR make the guard obsolete. You can just drop my commit https://github.com/Arduino-CI/arduino_ci/pull/183/commits/7d7d8103b877f6d9ddda84552a95ce3097ade3c3 completely.

ianfixes:
A squashed version of this was merged with #183

Issue #192, opened 2020-11-02
Arduino CI seems to not handle dependencies between imported Arduino Libraries


cherbel:
I am writing tests for a library A (TankControllerLib), and it imports library B (MAX31865) that has a dependance on library C (BusIO). When I wrote tests for library B, I included library C in .arduino-ci.yml, and it was able to find and use my modified library C. Now, I've included B and C as prerequisites for library A, and when library B is compiled, the header for library C is not found. 

```
In file included from /Users/c/Desktop/Fall 2020/CPTR 450 Software Engineering/Projects/TankControllerLib/src/Devices/TempProbe_TC.cpp:5:
In file included from /Users/c/Documents/Arduino/libraries/Adafruit_MAX31865/src/Adafruit_MAX31865_CI.h:2:
/Users/c/Documents/Arduino/libraries/Adafruit_MAX31865/src/Adafruit_MAX31865.h:63:10: fatal error: 'Adafruit_SPIDevice.h' file not found
#include <Adafruit_SPIDevice.h>
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Here is my .arduino_ci.yml for the TankControllerLib:
```
unittest:
    platforms:
        - mega2560
    libraries:
      - "LiquidCrystal"
      - "Adafruit BusIO"
      - "Adafruit_MAX31865"
```

Here is my .arduino_ci.yml for MAX31865:
```
unittest:
  platforms:
    - mega2560
  testfiles:
    reject:
      - "Common.cpp"
  libraries:
    - "Adafruit BusIO"
```

Issue report prepared with @jgfoster

ianfixes:
This might be fixed by my #183 branch, I added some recursive logic that leverages the `library.properties` spec.  Can you try the instructions in the PR description for how to test that out, and let me know if that works?

Kavikick:
I'm having the same problem with a different library, and the branch "2020-10-16_suggestions" does not seem to fix it.

ianfixes:
I should have linked to the [`library.properties` spec](https://arduino.github.io/arduino-cli/library-specification/).  And from you, it would be helpful to see the compiler invocation that produced the error -- it should be printed to the console above that error output.

[...]

Issue #193, opened 2020-11-04
Undeclared identifiers `digitalPinToPort` and `digitalPinToBitMask`


jgfoster:
When adding Arduino CI to [Adafruit_BusIO](https://github.com/Arduino-CI/Adafruit_BusIO) we get the following output:

```
Located Arduino binary...                              /Applications/Arduino.app
The set of compilers (1) isn't empty...                                        ✓
Checking g++ version... 
    Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
    Apple clang version 12.0.0 (clang-1200.0.32.21)
    Target: x86_64-apple-darwin19.6.0
    Thread model: posix
    InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
...Checking g++ version                                                        ✓
libasan availability for g++...                                             true
Requested unittest platform 'mega2560' is defined in 'platforms' YML...        ✓
Library conforms to Arduino library specification...                         1.5
Unit testing test.cpp with g++... 

[...]

PR #212, opened 2020-11-15
Add test for #192: fatal error: 'Adafruit_SPIDevice.h' file not found

jgfoster:
This should make it easier to debug #192.


ianfixes:
The motivation behind this PR seems to be illustrating the problem fixed by #213 but this doesn't quite close the loop... this project doesn't get included in the CI config.

I'm wondering if https://github.com/Arduino-CI/arduino_ci/tree/master/SampleProjects/DependOnSomething could be adapted to perform this test.

jgfoster:
This is illustrating the problem reported in #192 (not #193) which is not yet fixed. It wasn't added to CI because it would cause the CI to fail. Once #192 is fixed then we can include it into the CI.

jgfoster:
Replace with PR to `tdd` branch.

ianfixes:
Rolled into #223 and merged

ianfixes:
I got an email that

> I find that #212 still fails so this doesn't seem to be fixed.

I don't see that comment here so I assume it was retracted.  Please open a new issue if the problem comes back, as #218 will have affected quite a few things

jgfoster:
Yes, I discovered that the problem was elsewhere.

PR #213, opened 2020-11-15
Fix typo in `aux_libraries` name

jgfoster:
See https://github.com/jgfoster/Adafruit_MAX31865/runs/1401861281?check_suite_focus=true for a failure.


ianfixes:
I feel like I unit tested this, but obviously not enough.  Another test should probably be added but let's fix the bug for now

PR #216, opened 2020-11-16
Add test for #192: fatal error: 'Adafruit_SPIDevice.h' file not found

jgfoster:
Replace #212.


jgfoster:
This will be clearer if it follows #217.

PR #224, opened 2020-11-25
Fix library parse error if there is an extra blank line in `library_properties.rb`

jgfoster:
A library I'm testing had a blank line at the end of `library_properties.rb` and this caused an error. To see the error revert `lib/arduino_ci/library_properties.rb` and try running the tests.


ianfixes:
It looks like #218 and #225 contain everything that's in here

PR #225, opened 2020-11-26
Add `IPAddress.h` to `Client.h`

jgfoster:
This fixes a compile error when `Arduino.h` was not included first (as shown in the new test).


ianfixes:
I think this needs a rebase and then it's all set

PR #284, opened 2021-02-15
Test on push (not just on pull request)

jgfoster:
I found that when I made a change and pushed to my fork the tests were not run. I'd like to have them run so I can confirm my work before I submit a Pull Request. This addresses that need.


ianfixes:
In my local branch, I had enabled "Test on push" for just the Linux runner, is that sufficient?  I can also make the change in the documentation ("produces" -> "should produce") if so.

jgfoster:
Yes, Linux would be sufficient.

jgfoster:
@prestoncarman, would you give a review?

ianfixes (approved):
Re-reading the GitHub docs about billing, all open-source repositories seem to be allowed no-cost use of GitHub Actions... so there is no financial impact to this change (which was my original reason for not testing both pushes and pull requests)

prestoncarman (commented):
LGTM

PR #290, opened 2021-02-23
Hide build artifacts

jgfoster:
Put build artifacts in a separate directory to reduce clutter (the name could be anything).


ianfixes:
I kept the build artifacts around because I needed to do some `gdb` debugging of the test macros I wrote.  But that doesn't seem to be a very valid use case overall.

Can you think of any situations in which you'd actually want to keep those build artifacts around?  Because my thinking now is to just delete them after they are used, and _maybe later_ offer a command line option to keep them if they are really desired.

jgfoster:
I don't have any reason to keep them around, but if they are hidden then I don't care! But, what is meant by "after they are used"? I think it would be mildly distracting to have them appear and disappear during the build/test process. So, even if they were to be removed some time later, having them hidden during the whole process is a marginal preference.

ianfixes:
OK.  I will see about saving them as tempfiles instead.

PR #293, opened 2021-03-21
Support delay() callback/handler

jgfoster:
When running on the real Arduino, the `delay()` function will actually pause. When running on the mock Arduino, the `delay()` function increments the clock and then returns immediately. While this is generally the right thing to do, this presents a couple issues. First, if we want to look at the state during the delay, this is difficult since the state could change after the delay (e.g., show a message on the LCD for one second, then clear it). Second, if we are using the mock Arduino library to emulate the application in a GUI, we want to have a real delay. To solve this, we will allow code (such as tests and the GUI) to register to be notified of delays. Then we can take whatever actionion is appropriate.

To be notified of a delay, define a void function that accepts a single `unsigned long int` as a parameter and register the function using `addDelayHandler()`. You can remove the callback with `removeDelayHandler()`. For an example of the usage, see `millis_micros_and_delay` in `SampleProjects/TestSomething/test/godmode.cpp`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+- Ability to use event handlers on a delay
 
 ### Changed
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -709,3 +709,21 @@ unittest(wire_basics) {
   assertEqual(0, mosi->size());
 }
 ```
+
+### Delay
+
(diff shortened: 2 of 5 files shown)


jgfoster:
For the record, this proposal was based on an approach I took to solving the problems described above. Since then we've taken a different approach that involves a "wait state" and no direct calls to `delay()`. So, even if provided I wouldn't use this feature. It does still seem like a reasonably useful implementation of the "observer" pattern, but I could see an argument for waiting till it had a real use case. So, as always, comments welcome.

jgfoster:
Since I've taken another approach with this, I'm going to close this unmerged.

PR #294, opened 2021-03-24
Build a shared library for use with each test

### LINKED ISSUE #227: Compile `arduino_ci` once as a linkable library and use for each test
Each test is compiled with the mocks as a single compile/link operation. When there are many tests, this can take a bit of time. One possible optimization would be to compile `arduino_ci` once as a linkable library at the beginning and then link it with each subsequent test (which would compile more quickly).

In addition to the performance improvement, this would allow us to separate the process of building the mocks from the process of building the tests. In particular, this would allow us to use some C++ libraries and features when building the mocks that are not available when building the tests. That is, we want the tests to be built using an environment as close as possible to the Arduino CLI/IDE, but there isn't any reason (other than convenience) why we need to apply the same limits to the mocks. (This is a follow-up to #219.) (Linking different libraries from the same compiler [should be safe](https://stackoverflow.com/questions/46746878/is-it-safe-to-link-c17-c14-and-c11-objects).)

jgfoster:
Instead of recompiling all the source files for each test, we now build a single shared library with everything except the test and then compile the test and link to the shared library. When I test SampleProjects/TestSomething the time reduces from 2:25 to 0:48. On our TankControllerLib, tests typically take over 30 minutes on my machine and this change reduces the time to just over two minutes, so a massive win!

I realize that I'm getting pretty deep into the internals, and I've not tested this on Linux or Windows so we'll see if the GitHub Action catches anything!

Fix #227.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ vendor
 # C++ stuff
 *.bin
 *.bin.dSYM
+*.so
+*.so.dSYM
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 ### Changed
+- We now compile a shared library to be used for each test
 
 ### Deprecated
 
--- a/exe/arduino_ci.rb
+++ b/exe/arduino_ci.rb
@@ -416,9 +416,27 @@ def perform_unit_tests(cpp_library, file_config)
 
   platforms.each do |p|
     puts
-    config.allowable_unittest_files(cpp_library.test_files).each do |unittest_path|
-      unittest_name = unittest_path.basename.to_s
-      compilers.each do |gcc_binary|
+    compilers.each do |gcc_binary|
+      # before compiling the tests, build a shared library of everything except the test code
+      attempt_multiline("Build shared library with #{gcc_binary} for #{p}") do
+        exe = cpp_library.build_for_test_with_configuration(
+          nil, # nil is a flag that we are building the shared library with everything else
+          config.aux_libraries_for_unittest,
+          gcc_binary,
+          config.gcc_config(p)
+        )
+        puts
+        unless exe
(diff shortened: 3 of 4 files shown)


jgfoster:
With https://github.com/Open-Acidification/TankController/pull/207 our tests went from 42:30 to 2:30! (Unfortunately, we've had to start using a fork of `arduino_ci` in order to have this work.)

ianfixes:
I like what's going on here but I'd feel more comfortable if we found a way to factor this a bit better.  For example:

* a constant (or function) instead of a duplicated string to provide the name of the shared library
* a function to calculate the path of the shared library
* a standalone function responsible for building the library
* a standalone function to check whether the shared library exists (if necessary)

I can probably take a crack at this if you'd like

jgfoster:
As mentioned a moment ago in another PR, this is the remaining PR that I haven't merged and that is because I'm a bit hesitant, partly for the reasons you suggest (and partly because it doesn't give the speed improvements I anticipated). As long as this is unmerged, our TankController project will continue to use a fork (which is fine; we've been doing it for some months). 

So, yes, if you see what I'm trying to do and can find a way to factor it better, that would be wonderful. But, please don't feel an obligation to take my problems and try to solve them!

jgfoster:
@ianfixes, I've got this working and it is giving me the performance improvements I expected. A few days ago you mentioned a few ideas:

>a constant (or function) instead of a duplicated string to provide the name of the shared library

See `lib/arduino_ci/cpp_library.rb:494`.

>a function to calculate the path of the shared library

Would this replace the code at `lib/arduino_ci/cpp_library.rb:495`?

>a standalone function responsible for building the library

Is this to extract out the code at `exec/arduino_ci.rb:426-441` or is it to replace `build_for_test_with_configuration()` in `lib/arduino_ci/cpp.library.rb:491`?

>a standalone function to check whether the shared library exists (if necessary)

Would this be used at `lib/arduino_ci/cpp_library.rb:525` instead of `File.exists?(full_lib_name)`?

Issue #292, opened 2021-03-19
Suppress certain compiler warnings


jgfoster:
There are some compiler warnings in code unrelated to mine that I'd like to suppress. Changing `cpp_library.rb:497` to the following does what I want. Perhaps there could be a configuration in `.arduino-ci.yml` to which one could add warnings (like we have for defines)? I could create a PR if you think this is a good direction and approach.

```
      arg_sets << ["-std=c++11", "-o", executable.to_s, "-DARDUINO=100", "-Wno-unknown-attributes", "-Wno-address-of-packed-member"]
```

ianfixes:
`.arduino-ci.yml` does support this already, but at the platform level:

https://github.com/Arduino-CI/arduino_ci/blob/d06591aca30fb5cdcbb39fa726fe14f1100df71c/REFERENCE.md#L95-L109

It sounds like this PR wants to have a project-wide setting for suppressing warnings, which seems like a good idea.  I'd rather not hard-code it in `cpp_library.rb` at all.

jgfoster:
While it is generous of you to assume that I wanted something more than what was already provided, the truth is that I didn't know about the `warnings:` option and would be happy to have this closed with RTFM!

ianfixes:
It's up to you, I would be glad to learn more about the use case.  If there are a class of warnings that are generally unhelpful to you (i.e. not tied to a particular platform) then I'd be interested to learn more.  It does seems like we'd want to give the developers some more direct control here (i.e. without overriding platform definitions).

jgfoster:
I think there are enough other things that are higher priority!

PR #296, opened 2021-04-04
Big fix branch for 2021

ianfixes:
There are a lot of interrelated things to fix, so this branch intends to be a heavy-handed attempt to combine all those fixes.

## Highlights from `CHANGELOG.md`

* See CHANGELOG.md for more

## Issues Fixed

* Fixes #295
* Fixes #283
* Supersedes #285 
* Fixes #282
* Supersedes #297
* Fixes #288
* Supersedes #289
* Supersedes #298 
* Supersedes #286 
* Fixes #269
* Supersedes #290
* Addresses #280

PR #297, opened 2021-04-04
Fix problem with warnings for platforms

jgfoster:
As a follow-up to #292, I tried to add warnings for a platform and found that it doesn't work. This should fix it.


ianfixes:
Is the YAML file a necessary part of the fix?  If not, I'll just fix the typo in my branch

jgfoster:
The YAML file is a _test_ of the fix. If you add the YAML file without the `cpp_library.rb` then you can see the failure.

ianfixes:
Understood.  I'm going to exercise this with unit testing instead, as I should have done earlier to prevent such an obvious error.

ianfixes:
https://github.com/Arduino-CI/arduino_ci/pull/296/commits/375c3f3ea7e5215560c2cbcb66d3358821374666

PR #302, opened 2021-06-04
Stream::readStreamUntil() should not return delimiter

jgfoster:
The delimiter should not be included in the returned String.


ianfixes:
Did this turn out not to be a bug?

jgfoster:
Actually, no, it is a bug. I was doing some cleanup on one of my machines and accidentally deleted both the local and remote branches.

PR #306, opened 2021-08-19
Rule of three for Client

jgfoster:
if you have a custom destructor, then you probably need a custom copy constructor and a custom copy assignment operator.

### DIFF
--- a/SampleProjects/TestSomething/test/clientServer.cpp
+++ b/SampleProjects/TestSomething/test/clientServer.cpp
@@ -20,10 +20,13 @@ unittest(Client) {
 }
 
 unittest(Client_copy_constructor) {
-  Client client1;
-  Client client2;
-  client2 = client1;
-  assertTrue(true);
+  {                 // Client object contains a reference to a String object
+    Client c1;      // Constructor instantiates a String (s1)
+    Client c2;      // Constructor instantiates a String (s2)
+    c2 = c1;        // Does c2 get a reference to s1 or a copy of it?
+  }                 // End of scope calls destructor on c1 and c2
+  assertTrue(true); // Was s1 deleted once (with c1) or twice (also with c2)?
+                    // Was s2 deleted at all (should be during assignment)?
 }
 
 unittest(IPAddress) {
--- a/SampleProjects/TestSomething/test/clientServer.cpp
+++ b/SampleProjects/TestSomething/test/clientServer.cpp
@@ -19,6 +19,13 @@ unittest(Client) {
   assertEqual(outData + "\r\n", inData);
 }
 
+unittest(Client_copy_constructor) {
+  Client client1;
+  Client client2;
+  client2 = client1;
+  assertTrue(true);
+}
+
 unittest(IPAddress) {
   IPAddress ipAddress0;
   assertEqual(0, ipAddress0.asWord());
--- a/SampleProjects/TestSomething/test/clientServer.cpp
+++ b/SampleProjects/TestSomething/test/clientServer.cpp
@@ -20,13 +20,26 @@ unittest(Client) {
 }
 
 unittest(Client_copy_constructor) {
-  {                 // Client object contains a reference to a String object
-    Client c1;      // Constructor instantiates a String (s1)
-    Client c2;      // Constructor instantiates a String (s2)
-    c2 = c1;        // Does c2 get a reference to s1 or a copy of it?
-  }                 // End of scope calls destructor on c1 and c2
-  assertTrue(true); // Was s1 deleted once (with c1) or twice (also with c2)?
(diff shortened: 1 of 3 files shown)


SampleProjects/TestSomething/test/clientServer.cpp
```diff
@@ -19,6 +19,13 @@ unittest(Client) {
   assertEqual(outData + "\r\n", inData);
 }
 
+unittest(Client_copy_constructor) {
+  Client client1;
```
prestoncarman, line 23: Could you elaborate on what this test is doing?

SampleProjects/TestSomething/test/clientServer.cpp
```diff
@@ -20,10 +20,13 @@ unittest(Client) {
...
-  Client client1;
-  Client client2;
-  client2 = client1;
-  assertTrue(true);
+  {                 // Client object contains a reference to a String object
+    Client c1;      // Constructor instantiates a String (s1)
+    Client c2;      // Constructor instantiates a String (s2)
+    c2 = c1;        // Does c2 get a reference to s1 or a copy of it?
+  }                 // End of scope calls destructor on c1 and c2
+  assertTrue(true); // Was s1 deleted once (with c1) or twice (also with c2)?
```
ianfixes, line 28: Is it just me or does this comment not match the assertion (yet)?

jgfoster: Not just you! @prestoncarman thought it was still not very clear so I've updated the tests and comments. Let me know if the new code looks better!

prestoncarman (approved):
The test description and checks are more clear now.

SampleProjects/TestSomething/test/clientServer.cpp
```diff
@@ -19,6 +19,29 @@ unittest(Client) {
...
+    assertEqual("1", *(c1.mGodmodeDataIn));
+    assertEqual("1", *(c2.mGodmodeDataIn));
+    c1.write('1');
+    c2.write('2');
+    assertEqual("11", *(c1.mGodmodeDataIn));
+    assertEqual("12", *(c2.mGodmodeDataIn));
+  } // End of scope calls destructor on c1 and c2
+    // Memory monitoring will give an error if delete is called twice on string1
+    // The following assertion is just to confirm that we got through the above
+  assertTrue(true);
```
ianfixes, line 42: This `assertTrue(true);` is sticking out like a sore thumb for me, but with that comment added I understand the reasoning.

My only reservation here is that if tests can effectively crash without reporting failure status, that would seem like a bug.  If you have observed that behavior, can you open a bug to that effect?

jgfoster: While I'm willing to open a bug, a problem is that "it is impossible to catch SIGSEGV from within a program using strictly standard and portable C++. SIGSEGV is still caught by the operating system and is normally reported to the parent process when a wait family function is called." ([cite](https://stackoverflow.com/questions/2350489/how-to-catch-segmentation-fault-in-linux)). See [here](https://github.com/Arduino-CI/arduino_ci/pull/306/commits/23c3670b394d5f4ab4ffd6d31649f86607b5a3d7) for how the error is reported in our case. If you would like to see it for yourself, simply comment out the fix (`cpp/arduino/Client.h:14-32`) and run this test. While there are ways to catch a SIGSEGV, I don't see that the benefit would be worth the cost. The only real benefit is having the tests continue so we can learn if something else is broken. Since my typical practice is to fix one bug at a time (frequently subsequent errors are tied to the earlier errors), the value of the later reports is minimal.

ianfixes: Sorry, I should be more clear.  I don't think there is anything to do on the C++ side.  I'm looking here:
https://github.com/Arduino-CI/arduino_ci/blob/9369dac10df466df1f97ce57e0db3e099b287992/lib/arduino_ci/cpp_library.rb#L535-L542

The test runner expects every unit test executable to terminate successfully (exit code of `0`), with any actual unit test failures communicated via the [TAP-formatted output](https://testanything.org/).  So a segfault _should_ cause a nonzero exit code and the ruby test runner _should_ mark the unit test step as a failure, with the code as currently written.

jgfoster: I wasn't thinking of the Ruby code! Yes, I'll enter a bug for this. Otherwise, is this ready to merge?

ianfixes: I thought I had replied, I'm sorry to have kept you waiting.  I see that you opened https://github.com/Arduino-CI/arduino_ci/issues/312 as well -- thank you!

PR #308, opened 2021-08-27
Test for minimum free memory

### LINKED ISSUE #307: Support for memory usage testing
Because of the limited memory on an Arduino board, a common problem is stack/heap corruption and this is difficult to debug. One way to address the problem is to ensure a free space threshold. We do get the free space when the sketch is compiled and it would be nice to be able to test for that value (or at least report it).

jgfoster:
Fix #307.

### DIFF
--- a/exe/arduino_ci.rb
+++ b/exe/arduino_ci.rb
@@ -24,6 +24,7 @@ class Parser
       ci_config: {
         "unittest" => unit_config
       },
+      min_free_space: 0,
     }
 
     opt_parser = OptionParser.new do |opts|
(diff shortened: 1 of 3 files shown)


prestoncarman (approved):
LGTM

exe/arduino_ci.rb
```diff
@@ -478,8 +483,18 @@ def perform_example_compilation_tests(cpp_library, config)
...
+          output = @backend.last_msg
+          puts output
+          i = output.index("leaving")
+          free_space = output[i + 8..-1].to_i
+          min_free_space = @cli_options[:min_free_space]
+          if free_space < min_free_space
+            puts "Free space of #{free_space} is less than minimum of #{min_free_space}"
+            ret = false
+          end
+        else
```
ianfixes, line 497: Will clean this up in #335

PR #309, opened 2021-09-02
Support for dtostrf()

jgfoster:
This is a function provided by Arduino that isn't in the standard C library.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,40 @@
+# https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings
+# https://git-scm.com/docs/gitattributes
+# https://git-scm.com/docs/git-config 
+# https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/ 
+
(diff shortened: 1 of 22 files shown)


jgfoster:
Sorry, this should have been a branch from master. Let me know if you want it resubmitted.

jgfoster:
This PR was a branch from main instead of master so has been replaced with #309.

ianfixes:
Which is the replacement PR?  309 is _this_ issue

jgfoster:
Yes, that's right. My `main` has all my changes merged into it and I should be making my PRs on a branch from `master` which tracks upstream master. I made the mistake of branching from my `main` so the initial PR had a huge number of modified files. In order to do a clean merge I recreated the proposed changes in a clean PR.

jgfoster:
And, if you have any substantive comments on closed PRs (like this one), I'm happy to revert or supplement...

ianfixes:
Any changes that have unit tests attached are generally fine with me, we can always add to the suite if/when we find a problem with the implementation.

PR #317, opened 2021-10-26
Super linter

### LINKED ISSUE #310: Add Super Linter
@prestoncarman, Would you be able to do this?

https://github.com/github/super-linter

prestoncarman:

* Added a CI workflow to lint the code base

## Issues Fixed

* Fixes #310


prestoncarman:
I have found the GitHub Action called Super-Linter easy to set up and configure. So instead of creating individual actions, I thought we could start with using this tool until it does not meet our needs. These are the individual linters that are being run to cover C++, Markdown, and YAML files.

* cpp-lint
* clang-format
* markdownlint
* YamlLint

This PR only includes linter configuration and any conflicting linting issues that need to be fixed. The linting configuration files for each language have been configured to allow the current code to pass all the tests. In the future, the configuration settings could be reviewed to see if any of those settings should be changed.

README.md
```diff
@@ -120,15 +120,15 @@ gem 'arduino_ci', path: '/path/to/development/dir/for/arduino_ci'
 ### Installing the Dependencies
 
 Fulfilling the `arduino_ci` library dependency is as easy as running either of these two commands:
-```
+
+```sh
```
ianfixes, line 124: This should be `console`, not `sh`, because this code block includes the prompt characters.  The difference in handling is easier to see when the prompt itself has some formatting:

Code block with `sh` formatting
```sh
user@host:~$ bundle install   # adds packages to global library (may require admin rights)
user@host:~$ bundle install --path vendor/bundle   # adds packages to local library
```

Code block with `console` formatting
```console
user@host:~$ bundle install   # adds packages to global library (may require admin rights)
user@host:~$ bundle install --path vendor/bundle   # adds packages to local library
```

REFERENCE.md
```diff
@@ -273,7 +273,7 @@ unittest_main()
 ```
 
 
-# Build Scripts
+## Build Scripts
```
ianfixes, line 276: I'm not sure I agree with the linter on this.  "Build scripts" and "Mocks of Arduino Hardware Functions" are major sections in this document, which is why the titles were formatted as such.

SampleProjects/DoSomething/README.md
```diff
@@ -3,13 +3,13 @@
 This project is a template for a CI-enabled (and unit testable) Arduino project of your own.
 
 
-### Features
+## Features
```
ianfixes, line 6: Here again, the linter is doing something incorrect. The `###` title is by design, because it does not produce a horizontal rule when rendered -- an undesirable division within the top section of this document.

cpp/arduino/avr/CPPLINT.cfg
```diff
@@ -0,0 +1 @@
+exclude_files=.*\.h
```
ianfixes, line 1: If this config file format allows comments, can you add some explicit text to describe what is being ignored here & why?

README.md
prestoncarman: Nice, I did not know about the console option. Looks good.

cpp/arduino/avr/CPPLINT.cfg
prestoncarman: Yes, that would be helpful.

SampleProjects/DoSomething/README.md
prestoncarman: Here the linter wanted to increment the heading by one instead of two. I updated the linter to ignore this.

REFERENCE.md
prestoncarman: The linter suggests to only have one top level header. I removed the changes and updated the configuration.

prestoncarman (commented):
Thanks for the feedback. It appears I make a few unnecessary changes for linting. The rules have been updated.

Issue #291, opened 2021-03-10
Attempting to download Arduino 0.13.0 package fails


jgfoster:
When setting up a new machine without the Arduino IDE or the Arduino CLI, we get the following error.

```
Fetching gem metadata from https://rubygems.org/.....
Resolving dependencies...
Fetching os 1.1.1
Installing os 1.1.1
Fetching rubyzip 1.3.0
Installing rubyzip 1.3.0
Fetching arduino_ci 1.3.0
Installing arduino_ci 1.3.0
Using bundler 2.1.4
Bundle complete! 1 Gemfile dependency, 4 gems now installed.
Bundled gems are installed into `./vendor/bundle`
                           .                  __  ___
                   _, ,_  _| , . * ._   _    /  `  | 
                  (_| [ `(_] (_| | [ ) (_)   \__. _|_   v1.3.0

 

Host OS...                                                                 linux
Attempting to download Arduino 0.13.0 package with open-uriTraceback (most recent call last):
    10: from /home/carmpr/Documents/workspace-wwu/TankControllerLib/vendor/bundle/ruby/2.7.0/bin/arduino_ci.rb:23:in `<main>'
     9: from /home/carmpr/Documents/workspace-wwu/TankControllerLib/vendor/bundle/ruby/2.7.0/bin/arduino_ci.rb:23:in `load'
     8: from /home/carmpr/Documents/workspace-wwu/TankControllerLib/vendor/bundle/ruby/2.7.0/gems/arduino_ci-1.3.0/exe/arduino_ci.rb:497:in `<top (required)>'
     7: from /home/carmpr/Documents/workspace-wwu/TankControllerLib/vendor/bundle/ruby/2.7.0/gems/arduino_ci-1.3.0/lib/arduino_ci/arduino_installation.rb:43:in `autolocate!'
     6: from /home/carmpr/Documents/workspace-wwu/TankControllerLib/vendor/bundle/ruby/2.7.0/gems/arduino_ci-1.3.0/lib/arduino_ci/arduino_installation.rb:57:in `force_install'
     5: from /home/carmpr/Documents/workspace-wwu/TankControllerLib/vendor/bundle/ruby/2.7.0/gems/arduino_ci-1.3.0/lib/arduino_ci/arduino_downloader.rb:136:in `execute'
[...]

PR #319, opened 2022-01-27
Fix open-uri call for downloading release files

hlovdal:
This fixes the [issue](https://github.com/Arduino-CI/arduino_ci/issues/291) where attempting to download releases from <https://github.com/arduino/arduino-cli/releases/download/> fails.


jgfoster:
Very nice, thanks!

PR #326, opened 2022-07-01
Run "TestSomething" on Windows as well as Ubuntu

jgfoster:
Windows tests should be run on Windows not Ubuntu.


ianfixes:
Good catch

PR #327, opened 2022-07-01
Report compile errors in shared library

jgfoster:
Add test for compile errors.


exe/arduino_ci.rb
```diff
@@ -423,7 +423,22 @@ def perform_unit_tests(cpp_library, file_config)
...
+        )
+        unless exe
+          puts "Last command: #{cpp_library.last_cmd}"
+          puts cpp_library.last_out
+          puts cpp_library.last_err
+          got_shared_library = false
+        end
+        next got_shared_library
+      end
+      next unless got_shared_library
```
ianfixes, line 441: Can you explain this change a bit?  I'm not opposed to it, it just seems weird to move a small function into a big function

jgfoster: To be honest, I don't understand it myself. But the symptom was that a failure in the compile shared library function (which called the `attempt_multiline()` method) did not get properly reported out when it failed. That is, we got a compile error in the shared library then went on to compile the examples and when the examples compiled okay the overall test reported as passed. It seemed to me that when the `attempt_multiline()` method recognized a failure it reported it out one level but it didn't get out two levels. That is, by putting the call into a small, well-named method (which would otherwise be a good practice), it swallowed the error. But by moving it up a level the error got properly reported. 

Of course, this is all speculation and since I don't really understand what is going on I'm not real comfortable with the explanation. In an case, I added tests that confirmed the problem and verified the fix. But I really would welcome a better understanding of what changed and how this fixed it!

exe/arduino_ci.rb
```diff
@@ -445,24 +460,6 @@ def perform_unit_tests(cpp_library, file_config)
...
-      config.gcc_config(platform)
-    )
-    puts
-    unless exe
-      puts "Last command: #{cpp_library.last_cmd}"
-      puts cpp_library.last_out
-      puts cpp_library.last_err
-      return false
-    end
-    return true
```
ianfixes, line 462: @jgfoster I think the actual problem was this `return true` which should just be an implicit return of `exe` 🤦   I'll push up this change in a bit and see what happens

PR #338, opened 2022-12-28
Further bugfixes and documentation fixes

ianfixes:
This cleans up the `v1.4.0` release and makes progress on past milestones

* Fixes #159
* Fixes #339
* Fixes #336 
* Fixes #273
* Fixes #272

PR #342, opened 2022-12-29
Fix false positive result where test passed even though gcc failed with error

hlovdal:
Even if gcc failed with errors for some of the files compiled for unit test, the test would still be reported as successful. This is fixed by properly increasing `@failure_count`.

Also an additional fix for the result calculation for `all_arduino_library_dependencies!`. It is a long time since I made that commit so I do not exactly remember the details but reading it now accumulating with `[n]` inside the `map` section looks more correct.

--------

Updates to CHANGELOG.md will trigger conflicts for this branch (`failure_handling.m`, rebased on top of `master`) when #338 is completed (or if this is done first, then #338 will get conflicts).

There is an alternative branch `failure_handling` applied in series with other branches that does not have those conflicts when #338 is completed. Let me know if you rather want that one merged (later).


ianfixes:
Can you point me to an example of where this is happening and/or how I can reproduce it locally?

Also, I'm not clear on how the change to `cpp_library.rb` relates to the fix here.

lib/arduino_ci/cpp_library.rb
```diff
@@ -410,9 +410,9 @@ def all_arduino_library_dependencies!(additional_libraries = [])
       recursive = (additional_libraries + arduino_library_dependencies).map do |n|
         other_lib = self.class.new(n, @backend)
         other_lib.install unless other_lib.installed?
-        other_lib.all_arduino_library_dependencies!
+        [n] + other_lib.all_arduino_library_dependencies!
       end.flatten
-      (additional_libraries + recursive).uniq
+      recursive.uniq
     end
```
ianfixes, line 416: Do you have a test case or some performance metric that relates to this change?  Just trying to figure out whether this is an improved implementation or a bugfix

ianfixes:
I cherry-picked 5df7166b46b15b17baac73e743f800280fc9db5b into #338 but I'm reluctant to change the "additional libraries" recursive function without a unit test in place.

PR #350, opened 2023-03-15
Add util/atomic.h 

ghaug:
Hi,
I had the problem that I wanted to check a library modified by me which in turn includes a library that includes util/atomic.h and that doesn't work, of course. I added a mocked version of util/atomic.h, that simply does nothing. Atomicity by switching off interrupts is not possible on the host OS but timing behavior will differ anyway, therefore I think this approach is acceptable.
Best,
Gunter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+- Add util/atomic.h
 
 ### Changed
 
--- a/SampleProjects/TestSomething/test/atomic.cpp
+++ b/SampleProjects/TestSomething/test/atomic.cpp
@@ -1,0 +1,39 @@
+#include <ArduinoUnitTests.h>
+#include <Arduino.h>
+#include <util/atomic.h>
+
+
+unittest(atomic)
+{
+  // The macros don't do anything on the host platform, just make sure
+  // they compile without error.
+
+  int a = 1;
+  int b = 2;
+
+  ATOMIC_BLOCK(ATOMIC_RESTORESTATE) {
+    a += b;
+    b++;
+  }
+
+  ATOMIC_BLOCK(ATOMIC_FORCEON) {
+    a += b;
+    b++;
+  }
+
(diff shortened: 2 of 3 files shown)


ianfixes:
Hi Gunter, thanks for submitting this!  

Can you add (or just post here) a dummy/example usage of some `atomic.h` functions that we could refer to in a unit test?  That way, we have a guard against future edits breaking things unexpectedly.

ghaug:
Hi Ian,
thanks for the quick reply! There are two macros defined in atomic.h that can be used, namely ATOMIC_BLOCK and NONATOMIC_BLOCK. Each uses one of two parameters. An example would be:

`ATOMIC_BLOCK(ATOMIC_RESTORESTATE) {`
`    int a = callSomeFunction(para);`
`    passToSomeOtherFunktion(a);`
`}`

ATOMIC_BLOCK takes either ATOMIC_RESTORESTATE or ATOMIC_FORCEON as parameter, while NONATOMIC_BLOCK takes either NONATOMIC_RESTORESTATE or NONATOMIC_FORCEOFF, so there are just four combinations in total.

One more thing: Two of the checks fail for my pull request. I'm new to this project and I don't really know what's wrong.

ianfixes:
Regarding the test failures, I think you can ignore the one with the linter -- it looks like a temporary problem with GitHub (and if not, I can look into it later).  The only problem I'm seeing is a spelling error: https://github.com/Arduino-CI/arduino_ci/actions/runs/4424323994/jobs/7765538401?pr=350#step:4:19

Regarding a unit test, that would be awesome!  Please call it something like `atomic.cpp` and put it in [`SampleProjects/TestSomething/test`](https://github.com/Arduino-CI/arduino_ci/tree/master/SampleProjects/TestSomething/test).  For the format, just follow an example like [the minimal ones shown in `null.cpp`](https://github.com/Arduino-CI/arduino_ci/blob/master/SampleProjects/TestSomething/test/null.cpp) or check the docs: https://github.com/Arduino-CI/arduino_ci/blob/master/REFERENCE.md#most-basic-unit-test

ghaug:
I've added a unit test. As the macros don't do anything on the host, it just checks that all the four combinations compile without errors.

ianfixes:
This is excellent, thank you!

PR #355, opened 2023-07-18
Allow user-defined Makefile to optimize compiling

jgfoster:
Our [project](https://github.com/Open-Acidification/TankController) has about 100 application source files and 50 test files. When I repeatedly run the tests (with no changes) on my local machine they take about 3-1/2 minutes. About half of that time is spent determining what files to compile and what include directories to use and the other half of that time is spent doing the compiling (one file at a time). 

When (re)building a large project programmers typically use a tool like `make` to keep track of dependencies and avoid recompiling files that have not changed. Of course, there is a cost to maintaining the `Makefile` but it is typically worth it.

With this change we allow the user to provide a `Makefile` that considerably shortens the process (my project takes 10 seconds to rerun all tests if there are no changes; 15 seconds if there is a change to one test file; and 25 seconds if there is a change to an application file).

If the user does not provide a `Makefile` then the prior behavior is preserved. Note that this is disabled on Windows (I don't know enough about the build environment; does it support `g++`?).


jgfoster:
@Ian, We've been using this for a number of months based on my fork and I'd like to pull it into here so we can return to using this upstream project. Other than adding a test for the new behavior, the substance of the change is at lib/arduino_ci/cpp_library.rb:543-548. Let me know if you have any concerns; if not, I'll go ahead and merge it.

